PR #3, opened 2014-12-08
[ADD] l10n_fr_naf_ape from community-data-files

### LINKED ISSUE #1: Migrate l10n modules from community-data-files
The project https://github.com/OCA/community-data-files currently contains 2 modules that are localization-related, so it has been suggested they should be moved to other existing l10n_\* projects.

clonedagain:
Module moved from the OCA repository community-data-files.
First step to fix #1 .
The module works as-is on v8. README file added.


clonedagain:
Updated to include the revision history of the 2 modules which I had neglected to include.

yvaucher:
:+1:

legalsylvain:
Hi,
Thanks for porting this module. 2 thinks :
- It's better to make 2 PR for two modules;
- For l10n_eu_nace, I don't see the reason to include an european module into fr localization;
  
  :question:  CC : @alexis-via

yvaucher:
@legalsylvain Do you suggest to add an european repository?

legalsylvain:
In fact, if it is a EU project, it will be usefull for german, italian teams.
So, I would like to avoid forking that module into all eu localization. 
Other solution is to let this module into community-data-files, it isn't ?
What do you think ?
CC : OCA: @sebastienbeau, @gurneyalex, @alexis-via  ...

clonedagain:
The French part extends the EU part so I'd rather have both in the same 
repo, otherwise the French part is not installable.
I have no idea what's best, opinions are welcome.

__unported__/l10n_eu_nace/__openerp__.py
```diff
@@ -0,0 +1,54 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": "European NACE partner categories",
+    "version": "2.0",
+    'author': u'Numérigraphe SARL, Sistheo',
```
gurneyalex: Could you please add "Odoo Community Association (OCA)" to the list of authors?

__unported__/l10n_eu_nace/__openerp__.py
```diff
@@ -0,0 +1,54 @@
...
+from the RAMON service:
+    http://ec.europa.eu/eurostat/ramon/nomenclatures/index.cfm?TargetUrl=LST_CLS_DLD&StrNom=NACE_REV2&StrLanguageCode=DE&StrLayoutCode=#
+
+If you want to update the data or add another translation, download the
+corresponding file from RAMON using ',' as a separator, save it
+to the directory "data" and name it according to the language code:
+    NACE_REV2_<language code>.csv
+Then update the LANGS constant in the script "make_data.py" and run it to
+refresh the OpenERP data files. Finally, upgrade the module to load the data.
+""",
```
gurneyalex: I suggest moving the description to README.rst and using the OCA template.

__unported__/l10n_fr_naf_ape/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'French NAF partner categories and APE code',
+    'version': '1.0',
+    'author': u'Numérigraphe SARL',
```
gurneyalex: Could you add "Odoo Community Association (OCA)" to the list of authors?

__unported__/l10n_eu_nace/__openerp__.py
yvaucher: For now those modules are in `__unported__`

gurneyalex:
maybe we could use https://github.com/OCA/community-data-files for the `l10n_eu_nace` addon?

pedrobaeza:
Isn't already there but unported (https://github.com/OCA/community-data-files/tree/8.0/__unported__/l10n_eu_nace)? I think that one should be "OCA level", not localization level.

clonedagain:
Thanks all, I'll update this proposal based on your suggestions.

`l10n_eu_nace` can live on in `community-data-files` if it fits more there. I'll add a warning (EDIT: in the description) so interested users can install it.

clonedagain:
Done. I've also tested the modules on v8: they work unchanged.

Travis fails here because the `community-data-files` part is `__unported__`.
Can someone please review https://github.com/OCA/community-data-files/pull/3 ?

l10n_fr_naf_ape/README.rst
```diff
@@ -0,0 +1,46 @@
...
+(official main activity of the company), picked among the NAF nomenclature.
+
+Installation
+============
+Because the NAF is based on the NACE of the European Union, this module
+depends on the categories provided by the module `l10n_eu_nace`.
+To install this module, you may either:
+
+* download it from the Odoo app store: https://www.odoo.com/apps?search=l10n_eu_nace
+* OR download and install the code of the OCA project *Community data files*: https://github.com/numerigraphe/community-data-files
```
gurneyalex: could you use https://github.com/OCA/community-data-files ?

l10n_fr_naf_ape/partner.py
```diff
@@ -0,0 +1,34 @@
...
+from openerp.osv import osv, fields
+
+
+class Partner(osv.Model):
+    """Add the French APE (official main activity of the company)"""
+    _inherit = 'res.partner'
+    _columns = {
+        'ape_id': fields.many2one(
+            'res.partner.category', 'APE',
+            help="If the partner is a French company, enter it's official "
```
gurneyalex: s/it's/its/

l10n_fr_naf_ape/partner.py
```diff
@@ -0,0 +1,34 @@
...
+                 "main activity in this field. The APE is chosen among the "
+                 "NAF nomenclature."),
+    }
```
gurneyalex, line 34: question: I'm wondering if this ape_id should be added to the fields returned by res_partner._commercial_fields() to be copied to contacts (cf. https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/res/res_partner.py#L446)

gurneyalex:
overall :+1: once points above have been resolved

legalsylvain:
> Thanks all, I'll update this proposal based on your suggestions.
> l10n_eu_nace can live on in community-data-files if it fits more there. I'll add a warning (EDIT: in the description) so interested users can install it.

@clonedagain : Just my PoV, but I think the warning is unecessary. All dependencies are written in Travis File. So information is redundant. It should be usefull if it was a non OCA repository.

legalsylvain:
:+1: Code review. No test.
Thanks for porting this module and for the refactoring into to various repositories.

yvaucher:
needs fixing as some remarks from @gurneyalex still needs to be addressed

l10n_fr_naf_ape/partner.py
clonedagain, line 34: At the moment I don't think so, since the field is hidden for non-company contacts. But if someone has a use for it I wouldn't block their PR.

clonedagain:
Fixed I think, Travis should succeed once OCA/community-data-files#3 gets merged.

clonedagain:
Build passes if you force it: https://travis-ci.org/OCA/l10n-france/builds/54259385
Can we merge please ?

yvaucher:
Travis restarted

yvaucher:
Ok merged

PR #4, opened 2014-12-08
[ADD] l10n_fr_siret

### LINKED ISSUE #2: Migrate l10n_fr_siret from account-financial-tools
l10n_fr_siret was accepted in account-financial-tools but now that we have this l10n repo, it would be better suited here.

clonedagain:
Modules moved from the OCA repository account-financial-tools.

First step to fix #2.
The modules are not actually ported to v8, so I moved them to unported just to make it clear this is where they should be.
We intend to migrate them shortly after.


yvaucher:
Don't we loose it's history there ?
:-1:

alexis-via:
@clonedagain please accept my PR to your branch https://github.com/numerigraphe/l10n-france/pull/1

@yvaucher Of course it's always better to keep the history, but for a small module with 30 lines of code and a single author, it's not worth bothering a GIT super-expert to try to move the module with it's revision history. I don't think we have a GIT super-expert who is active in the French localization... I think we just need to move forward and don't loose time on this.

clonedagain:
Hi,
Yes we'll lose history indeed, I forgot bout that.
I humbly think I could make it a git-fu exercice to make a branch 
keeping the history but is it worth it?

rvalyi:
Hello @alexis-via I can help you with that when you want, this is easy with git-subtree for instance, see http://makingsoftware.wordpress.com/2013/02/16/using-git-subtrees-for-repository-separation

I think this is time to be serious with these history things, because there will be people that will try to rewrite it if you let it go sadly. Some want to take advantage that the largest part of the eco-system is exactly made out of these small pieces that may look small taken individually, so we better be rigorous with all these small insignificant pieces or some bad rounding errors about community assets will be made...

yvaucher:
This might help:

You can use git filter-branch to keep only commit of your module and then merge this extracted module to this repo:

http://gbayer.com/development/moving-files-from-one-git-repository-to-another-preserving-history/

clonedagain:
No problem I'll dig into this and update this PR with the history and 
the proposed migration to v8, possibly in a few hours if I can make it.

clonedagain:
Now that looks better. The history goes back to v5 code as it should.
Filtered repo account-financial-tools  with `git filter-branch --subdirectory-filter /l10n_fr_siret/ -f`, and then cleanly fused following the advices on this page: https://help.github.com/articles/about-git-subtree-merges/

I've merged in @alexis-via 's migration to v8, so this is not `_unported_` and I don't consider myself the actual author of this PR anymore.

So please allow me to say :+1: for @alexis-via 's work.

yvaucher:
thanks for your efforts :+1:

alexis-via:
@clonedagain Thanks for taking the time to fight with git to get the full history.

guewen:
:+1:

PR #10, opened 2015-02-04
account_bank_statement_import_fr_cfonb: Port to the new bank import framework

alexis-via:
This update is required to make the module compatible with the bank import framework currently under review in this PR: https://github.com/OCA/bank-statement-import/pull/6


alexis-via:
The new bank account import framework has been merged in the OCA branch, so we MUST merge this now, because the current version doesn't work any more :+1:

PR #16, opened 2015-03-24
non-official version of the export, with unposted entries

mdietrichc2c:
This PR is to add a new option to the FEC wizard, "non-official", which creates a similar FEC file but with unposted entries. The goal is to allow exporting to other financial software which can use the FEC format.

The file is also renamed to mark that it is not the official report.


gurneyalex:
looks ok. Is is possible to have a test for this?

alexis-via:
code review :+1:

yvaucher:
code review :+1:

A test would be great

PR #17, opened 2015-03-25
import the external dependency locally

gurneyalex:
this will prevent a crash at odoo startup when an external dependency
for an uninstalled addon is not available.


alexis-via:
:+1:

bguillot:
:+1:

alexis-via:
@gurneyalex u can rebase and merge

PR #24, opened 2015-04-26
8.0 Port to new API

alexis-via:
Add YAML test in l10n_fr_department


l10n_fr_department/model/res_partner.py
```diff
@@ -20,56 +20,38 @@
...
+    @api.depends('zip', 'country_id')
+    # If a department code changes, it will have to be manually recomputed
+    def _compute_department(self):
         '''This method is designed to be inherited'''
         dpt_id = False
-        zip = partner.zip
+        zip = self.zip
+        print "self=", self
+        print "zip=", zip
+        print "code=", self.country_id.code
```
yvaucher: You certainly want to remove those prints

l10n_fr_department/model/res_partner.py
```diff
@@ -20,56 +20,38 @@
...
-                    ('country_id', '=', partner.country_id.id),
-                    ], context=context)
-            if len(dpt_ids) == 1:
-                dpt_id = dpt_ids[0]
-        return dpt_id
+            dpts = self.env['res.country.department'].search([
+                ('code', '=', code),
+                ('country_id', '=', self.country_id.id),
+                ])
+            print "dpts=", dpts
```
yvaucher: Another print to remove

l10n_fr_department/model/res_partner.py
```diff
@@ -20,56 +20,38 @@
...
+            if len(dpts) == 1:
+                dpt_id = dpts[0].id
```
yvaucher, line 47: If dpts len is 1 you can acces to id without [0]

l10n_fr_department/model/res_partner.py
```diff
@@ -20,56 +20,38 @@
...
+        print "dpt_id=", dpt_id
```
yvaucher: Print to erase

l10n_fr_department/model/res_partner.py
```diff
@@ -20,56 +20,38 @@
...
+from openerp import models, fields, api
 
 
-class res_partner(orm.Model):
+class ResPartner(models.Model):
     _inherit = 'res.partner'
 
-    def _get_departement_of_partner(self, cr, uid, partner, context=None):
+    @api.one
+    @api.depends('zip', 'country_id')
```
yvaucher: maybe check for 'country_id.code' instead of only 'country_id'

l10n_fr_fec/wizard/fec.py
```diff
@@ -20,47 +20,39 @@
...
-        cur_wiz = self.browse(cr, uid, ids[0], context=context)
-        period_ids = self.pool['account.period'].search(
-            cr, uid, [('fiscalyear_id', '=', cur_wiz.fiscalyear_id.id)],
-            context=context)
-        assert period_ids, 'The Fiscal Year must have periods'
+    @api.multi
+    def generate_fec(self):
+        self.ensure_one()
+        periods = self.env['account.period'].search(
+            [('fiscalyear_id', '=', self.fiscalyear_id.id)])
```
yvaucher: `account.fiscalyear` has a field `period_ids` that you could use.

l10n_fr_fec/wizard/fec.py
```diff
@@ -182,8 +171,7 @@ def generate_fec(self, cr, uid, ids, context=None):
             'type': 'ir.actions.act_window',
             'view_mode': 'form',
             'res_model': self._name,
-            'res_id': ids[0],
+            'res_id': self[0].id,
```
yvaucher: as you have an ensure_one you can write
`self.id`

yvaucher:
Nice work. See my few remarks for some small improvements.

l10n_fr_fec/wizard/fec.py
alexis-via: Yes, I know. I write it this way because I prefer to make it explicit that I take the first recordet, rather than use the implicit "shortcut".

yvaucher:
@alexis-via It needs a rebase

l10n_fr_fec/wizard/fec.py
yvaucher: I see record set as it can be 2 things in one. Making this clear with `ensure_one` here you could consider your self is a single record.
But ok this is a matter of taste.

bguillot:
:+1:  
It seems good now. Thanks

sebastienbeau:
Now it's ok for me :+1:

PR #25, opened 2015-05-13
8.0 Fix FEC file to make it valid according to the official validation tool

alexis-via:
The FEC file generated by Odoo is now valid according to this tool http://www.economie.gouv.fr/dgfip/outil-test-des-fichiers-des-ecritures-comptables-fec

It has been tested on a real-life Odoo v8 database that even contains '|' in account move line labels !


chafique-delli:
Thinks Alexis for your work. :+1:

yvaucher:
:+1:

PR #39, opened 2015-10-09
[8.0] port module account_balance_ebp_csv_export from 7.0 to 8.0

alexis-via:


alexis-via:
I still have to fix a bug about the "to_display" property. So it's not ready as I thought...

alexis-via:
In fact, the module for 7.0 is also fucked, because of a very bad error in the addons-path of my test odoo server on my laptop... sorry for that. I'll fix all this ASAP.

alexis-via:
I'll re-submit it when it's ready.

PR #41, opened 2015-10-12
8.0 LCR CFONB file: FIX possible error with cents

alexis-via:
One of my customers showed me a scenario that he encountered in real life which triggered a bug with the current code in the generation of the LCR file:
line.amount_total was 292.21
line.amount_total \* 100 was 29221.0
int(line.amount_total \* 100) was 29220
And then he had an error when uploading the file to the bank because the total amont of the last line was different from the sum of the amounts of the line.

This fix should avoid this bug.


MOSAMAD:
:+1:

brother-bernard:
:+1:  No problem !

PR #44, opened 2015-10-21
Better unique ID for CFONB bank statement import

alexis-via:
I found out that, with some banks such as Crédit du Nord, the 'ref' field is always '000000', which makes a weaks unique_id with the current code. So I made a more complete unique_id with date and label.


cdubuit:
:+1:

PR #61, opened 2016-02-01
[9.0]  [l10n_fr_siret]  new:add siret unicity

njeudy:
- Add siret unicity constraint.
- Shown final user company name that already has this siret number


l10n_fr_siret/partner.py
```diff
@@ -58,6 +57,12 @@ def _get_siret(self):
     @api.constrains('siren', 'nic')
     def _check_siret(self):
         """Check the SIREN's and NIC's keys (last digits)"""
+        # Check unicity of siret
+        for partner in self.env['res.partner'].search([('id', '!=', self.id),('siret', 'not in', [False,''])]):
+            siret = str(self.siren) + str(self.nic)
+            if siret == partner.siret:
+                raise ValidationError(_("Siret must be unique !\n It was defined for '%s'.") % partner.name)
+
```
gurneyalex, line 65: why not a UNIQUE SQL index on siren + nic ? I'm concerned about performance issues with this implementation.

njeudy, line 65: @gurneyalex you're right it could be better to use UNIQUE SQL index on big database .. will try it and update code accordingly.

gurneyalex:
@njeudy setting needs fixing for now.

alexis-via:
I agree that siret unicity is a good idea. But, if we have siret unicity as default, we should also have vat number unicity as defaut, but there is no such thing in the official addons nor in the OCA addons (at least I'm not aware of). So I don't know what to think about this. But, in the end, I tend to think that having siret unicity by default is a good thing in most scenarios... the problem is that it's not consistent with the handling of the vat field.

njeudy:
I think we have to clone this one and have a clean discussion on this ?

@alexis-via

njeudy:
@gurneyalex @alexis-via think we should close this merge request ? 

Is it a good idea to have siret unicity ?

njeudy:
Close it for now has I did not use V9 anymore .. will open an other if I need.

l10n_fr_siret/partner.py
alexis-via: Note that it is not possible to have a UNIQUE SQL index based on siren (or siren + nic) because, with the current code of l10n_fr_siret, siren + nic is a commercial field, so it is synced between parent and child partners, cf https://github.com/OCA/l10n-france/blob/10.0/l10n_fr_siret/models/partner.py#L87

I think it sucks, but it is the same behaviour as the VAT field, cf https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_partner.py#L411

I would prefer to avoid this sync and oblige developers to read the "commercial_partner" to get the vat or siren/siret... but it's difficult to change this.

PR #66, opened 2016-02-27
8.0 CFONB: make it work with bank statement files of LCL

alexis-via:
Give default values for the fields currency (EUR) and decimal (2), to accept LCL files that are not fully compliant. I also sent a bug report to LCL a few month ago, but we still don't have any change on their side.


SMLGP:
+1: using it in production and I confirm it solves the problem".

PR #67, opened 2016-03-01
[FIX] do not include debit=credit=0 move lines (invalid for FEC)

mdietrichc2c:
Forward-port of an old merge in 6.1: http://bazaar.launchpad.net/~camptocamp/openerp-french-localization/6.1-remove-null-move-lines-mdh/revision/4 , needed since move lines with debit=credit=0 are marked as invalid by the FEC validation software.

### DIFF
--- a/l10n_fr_fec/wizard/fec.py
+++ b/l10n_fr_fec/wizard/fec.py
@@ -126,6 +126,7 @@ class account_fr_fec(orm.TransientModel):
         WHERE
             am.period_id IN %s
             AND am.company_id = %s
+            AND (aml.debit != 0 OR aml.credit != 0)
         '''
         # For official report: only use posted entries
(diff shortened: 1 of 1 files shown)


fclementic2c:
:+1:

alexis-via:
yes, we have that in the v8 branch but not in the v7 branch. By the way, you could have kept the SQL syntax of the v8 branch.

:+1:

mdietrichc2c:
@alexis-via that's true ; I did not check the 8.0 version.
I modified the fix to do a backport of the 8.0 instead.

sbidoul:
:+1: 

BTW @alexis-via, we are currently updating mis_builder to handle null values properly (ie display blank cells when there are no moves in the period).

vrenaville:
:+1:

PR #63, opened 2016-02-05
Revert merge of OCA/l10n-france#47 that broke both travis and runbot

lepistone:
#47 added a dependency on aeroo, which probably we don't want (see #59) and broke both travis and runbot builds on this branch.

I suggest we revert quickly as this affects all other pull requests, and then we can discuss with @alexis-via if he wants to submit another version of the module.

Thanks to all!

PR #68, opened 2016-03-01
8.0 ebp export without aeroo

sebastienbeau:
Add an ebp export version without Aeroo to solve this issue : https://github.com/OCA/l10n-france/pull/63


account_balance_ebp_csv_export/README.rst
```diff
@@ -0,0 +1,43 @@
+Account Balance EBP CSV export
```
pedrobaeza, line 6: Please use latest OCA's README: https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

account_balance_ebp_csv_export/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+{
+    'name': 'Account Balance EBP CSV export',
+    'version': '8.0.0.0.1',
```
pedrobaeza: Use 8.0.1.0.0

pedrobaeza:
A little remark, but :+1: (code review)

legalsylvain:
Is it still WIP ?

gurneyalex:
:+1:

alexis-via:
I found a bug: the filtering by period or by date is not taken into account in the EBP CSV export ; the amounts given are the amounts without any date/period filter.

The bug was also present in the old version of the module "with aeroo".

alexis-via:
FYI, I fixed the bug of the missing filter by date/period in the "aeroo" version of the module in this commit https://github.com/akretion/l10n-france/commit/5e8407c12e5ceeae274f76cc87a6bf5303fe17dc (it's in my branch 8.0-ebp-csv-import-aeroo where I keep the aeroo version for those that use it in production).

Also, it seems that the EBP CSV export should be encoded in latin1 and not in UTF-8. I made the change in the aeroo version and I'll have a confirmation when my customer will test it with Sage Etats Financiers.

sebastienbeau:
@alexis-via Ok I will update the export with the change done in aeroo version

alexis-via:
Seems to work nice with Teledec.fr.

alexis-via:
I took into account Pedro's remarks.

We used this version of the module to declare Akretion's liasse fiscale via teledec today and it worked very well. So :+1:  and I merge.

PR #70, opened 2016-03-03
Fill mandatory fields with default values

mdietrichc2c:
Another issue raised by a customer : after migrating from v5.0, some moves and move lines did not have set names, and the FEC software raised an issue.


vrenaville:
:+1:

alexis-via:
The 'name' field on account.move is required=True in v8, v7, v6.1, v6.0 and also v5.0, cf https://github.com/odoo/odoo/blob/5.0/addons/account/account.py#L906

I went back as far as December 2006, and that field was already required=True, cf:
https://github.com/odoo/odoo/blob/f72ee06ba3e55ab92b43258571b9c95a1a47ef63/addons/account/account.py#L463

So could you explain us why the 'name' field is empty on some of the account.move of your customer ? Isn't it a problem with the data of your customer or with your migration script ?

Personnally, I would have fixed the data instead of the code...

alexis-via:
My previous comment is also true for the 'name' field of account.move.line, which is required = True in v8, v7, v6.1, v6.0 and v5.
As this requied=True is defined in the Python code, Odoo should have added the property "not null" on the postgresql field... I really don't understand how it is possible to have that field empty...

mdietrichc2c:
@alexis-via thanks for the review. It was probably something due to the migration (it is a very old customer ; I am not sure on what version they started, and what happened before), and I did fix the data manually.

However, there is another part to the fix : you can see I test for an empty string, but also for a string only containing blank spaces (which is a valid value for mandatory fields), since a "blank" string would also fail with the FEC software.
I'm not sure what to do if the string is "blank" ; I added a default value, but it could be argued that it's an invalid value on the moves/move lines, and should not be allowed as a valid name. What do you propose?

alexis-via:
OK, I get your point about spaces, etc... fine for me :+1:

legalsylvain:
hi @mdietrichc2c could you fix minors conventions bugs that break Travis ?

https://travis-ci.org/OCA/l10n-france/jobs/113356505

kind regards.

mdietrichc2c:
@legalsylvain I fixed my issues, then other issues, which then raised more issues... not sure if that was helpful.

PR #71, opened 2016-03-08
8.0 FEC: VAT numbers can contain spaces in Odoo

alexis-via:
Take into account special forms of companies that have SIRET but no VAT


fclementic2c:
:+1:

PR #75, opened 2016-06-05
New module account bank statement import fr multi cfonb

fmdl:


alexis-via:
Support for multi-account files is coming in account_statement_import in v14 !

PR #79, opened 2016-07-05
9.0 Remove module l10n_fr_tax_sale_ttc

alexis-via:
It seems to be impossible in v9 to use the module "l10n_fr_tax_sale_ttc": even if the module is installed at the same time as l10n_fr, the tax templates defined in "l10n_fr_tax_sale_ttc" are not loaded to create real account.tax. So I made a PR to integrate this in the official l10n_fr module, cf https://github.com/odoo/odoo/pull/12695/files


alexis-via:
Was a duplicate of pr #84

PR #80, opened 2016-07-07
9.0 Migration to new regions + add overseas regions and departments

alexis-via:
This is a forward port of the work carried out in v8.


l10n_fr_state/data/res_country_state_data.yml
```diff
@@ -23,173 +24,102 @@
   Creating a res.country.state record
 - 
   !record {model: res.country.state, id: res_country_state_alsace}:
-    code: A
+    code: 44
```
legalsylvain, line 9: Hi, 
What is this code (44 for Grand Est) ?

l10n_fr_state/migrations/9.0.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,61 @@
...
+# © 2015-2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+def migrate(cr, version):
+    if not version:
+        return
+
+    # key : source region XMLID - value : destination region XMLID
+    new_region_map = {
+        'res_country_state_rhonealpes': 'res_country_state_auvergne',
```
legalsylvain, line 12: It should be great to have correct name for xml_id. For exemple, res_country_state_auvergne_rhone_alpes. It is less error prone.
You can make it easy with

```
from openupgradelib import openupgrade

xmlid_renames = [
     ('res_country_state_auvergne', 'res_country_state_auvergne_rhone_alpes'),
]

@openupgrade.migrate()
def migrate(cr, version):
    openupgrade.rename_xmlids(cr, xmlid_renames)
```

in the premigration script.

legalsylvain:
Hi,
some remarks (and some travis issues to fix).

Otherwise, LGTM. Thanks for this work !

fclementic2c:
:+1:

l10n_fr_state/data/res_country_state_data.yml
alexis-via, line 9: It's the code used by geonames.org. I want to use the same code as geonames.org to be able to use the module "OCA/partner-contact/base_location_geonames_import".

l10n_fr_state/migrations/9.0.0.1.1/pre-migration.py
alexis-via, line 12: I don't think that xmlid should be changed (at least not changed in a stable version) ; that's why the xmlid were left unchanged in v8 (this PR is just a forward-port of the changes in v8). xmlid is a technical name visible only to the developers ; if someone rely on the current xmlid, it will break his code, and we don't want that in a stable branch.

legalsylvain, line 12: You're right. We could do that when we'll port in V10. Better and more stable.

l10n_fr_state/data/res_country_state_data.yml
legalsylvain, line 9: Well, in fact, the current code is based on another norme (see description). It could be better to use another name 'geoname_code' (or something similar) to avoid to break current use and keep consistency. Could you add too a little text in the description file ? 

kind regard.

alexis-via, line 9: I added a note about this in README.

alexis-via:
I took into accounts all the remarks I think. So it should be ready to merge.

alexis-via:
I added overseas departments and regions, because we need to have them (and I need them in one of my projects) !

legalsylvain:
Hi, 

1) About the overseas department you added in the res.country.department model, they exists in the res.country table. (in Odoo & OCA core) I'm not in favor to have a database with guyane (and other) in two different tables. It will be error prone for the end user. 
https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/res/res_country_data.xml#L1235
I set a note here, when I wrote this module :  
https://github.com/OCA/l10n-france/blob/9.0/l10n_fr_state/__openerp__.py#L41

Maybe it's not the better classification, but editor has did that choice. If it's not the better, core should be fixed before.

 2) README file is missing in your PR or did I missed something ?

3) About code / geoname_code of the res.country.state model, you did'nt change the name of your field, so this PR is stilling breaking odoo norms.

Current state code for us in odoo : 
https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/res/res.country.state.csv#L10
that is the ISO_3166-2 classification. ref : https://fr.wikipedia.org/wiki/ISO_3166-2:US

So to be consistent, french state should use the classification : https://fr.wikipedia.org/wiki/ISO_3166-2:FR

Thanks.

alexis-via:
@legalsylvain About your last comment:
1) I think there are 2 incompatible datasets for this:
a) the ISO-3166 dataset, which give a country code to each DOM-TOM and doesn't give a department code to DOM and ignore the overseas regions
b) the legal dataset which says that there are 13+5 = 18 french régions and 95 + 5 = 100 french départments.

Initially, you decided to implement the ISO dataset. I need the legal dataset, both for some legal French declarations (intrastat in particular, but it may be needed elsewhere) and for some functional needs of my customer (I need the list of all French departments in a single table).
There is no magic solution: those 2 datasets are not compatible. So I propose the following:
- the l10n_fr_department module will contain the res.country.department table, but will not provide data
- for the data, we will have 2 sets of module: l10n_fr_state (which we will rename l10n_fr_state_iso in v10) and l10n_fr_department_iso that will implement the ISO dataset ; l10n_fr_state_legal and l10n_fr_department_legal that will implement the legal dataset. These modules will be flagged as incompatible.

2) I forgot the git add on the file, sorry ! This is now fixed.

3) Yes, we need to find a solution for geoname compatibility, but the geoname import module is a generic module, it is not specific to France. Maybe the solution is to put a hack in "l10n_fr_base_location_geonames_import" (which may required a modification in "base_location_geonames_import"). We need to find the right solution for this. This PR was just a forward port of the code in v8, where the ISO 3166 codes were dropped for regions. But I'm ok to try to find a solution to keep the ISO codes in v9.

alexis-via:
By the way, do you know where we can find the ISO3166 codes for the new French regions ? The wikipedia page is not up-to-date with the new regions of January 1st 2016.

alexis-via:
According to https://en.wikipedia.org/wiki/ISO_3166-2:FR , the ISO3166-2 codes have still not been updated for the new regions of January 1st 2016. If this is true, it means we can't use it for the moment... but this wikipedia page may not be up-to-date.

pedrobaeza:
We already have in Spain a hack in geonames import to change states codes to what we need: https://github.com/OCA/l10n-spain/blob/8.0/l10n_es_toponyms/wizard/geonames_import.py

alexis-via:
According to my research on the ISO website, the last version of 3166-2 is dated in 2013, so it is still not up-to-date with the new regions. So we can't use the ISO codes for the regions for the moment. But I'm not an expert user of the ISO website, so maybe I missed something ; @legalsylvain if you find the ISO codes for the new french regions, please give us the link.

alexis-via:
@pedrobaeza Thanks for your pointer ; this is a "clean" hack ; I guess we should do the same thing in "l10n_fr_base_location_geonames_import" !

alexis-via:
@legalsylvain 
Eventually, I decided to implement a lighter solution than the one I proposed yesterday: I just add the oversea states/departments via a single separate module l10n_fr_department_oversea. So:
a) those who want the ISO dataset will install only l10n_fr_state and/or l10n_fr_department
b) those who want the "legal" dataset will install l10n_fr_department_oversea (which depends on l10n_fr_department)

That way, we have a solution for everyone ! OK with this ?

About the nomenclature for regions, I now say in the README that we switch to the INSEE nomenclature for regions because the ISO 3166-2 nomenclature hasn't been updated following the new regions of 1/1/2016. If the ISO 3166-2 nomenclature is updated one day, we will be able to switch back from INSEE to ISO via a migration script. I think we don't have the choice on this for the moment.

sla-subteno-it:
👍

legalsylvain:
Hi, 

1) thanks a lot for this solution (_oversea module). This is elegant, and allow user to works with default odoo design (dom in res.country table) or in department table. (OCA design)
2) OK.
3) Odoo use 'code' field with iso 3166 nomenclature. I understand that it's maybe not the better nomenclature but to be consistent, OCA module should use same nomenclature. Otherwise, it's confusing. I don't understand why creating a new field 'xxx_code' is not possible.

PR #81, opened 2016-07-08
Add possibility to ignore accouns from cfonb files

florian-dacosta:
Give possibility to easily ignore some accounts from cfonb files in a custom module

2 goals : 
In Cfonb files, we may have multi account.
Sometimes, the second account is not used, but it is in the file anyway, which block the import. (as multi-account files are not managed in this module for now).

Also, we may want to import a specific account from the file, but not others.


alexis-via:
:+1: always good to have that kind of hooks.

PR #82, opened 2016-09-21
8.0 l10n_fr_department: small usability improvements

alexis-via:
Display department names like 'Rhône (69)' instead of just 'Rhône'

Add group by in search views


cdubuit:
:+1: merci

PR #86, opened 2016-10-06
10.0 Port modules for régions, departments, LCR, CFONB and SIRET

alexis-via:
Port 3 modules to v10:
l10n_fr_state
l10n_fr_department
l10n_fr_department_oversea


alexis-via:
Travis is green 3/4 ; the error on the last one is unrelated (it cannot download OCB 10)

alexis-via:
I also ported the modules for LCR and siret.

petrus-v:
:+1:

brother-bernard:
One important thing that came to my mid just now: we need to decide if we use the OCA module "partner_identification" in v10 https://github.com/OCA/partner-contact/tree/10.0/partner_identification to store SIREN/NIC or not. If we do, we'll drop the module l10n_fr_siret (or re-write it just to have a computed field that looks into the data of the partner_identification module). If not, then there is nothing to do, because I already ported l10n_fr_siret "as is" in this PR.
@clonedagain @sebastienbeau What's your opinion on this ?

alexis-via:
Oups, the previous comment was written by me, but with a web browser that had a session of Brother Bernard. Sorry.

clonedagain:
partner_identification looks just fine, a generic solution is always better I suppose.
l10n_fr_siret has checksum control I think : it's best if that can be kept but IMHO it should not block you otherwise.

alexis-via:
@clonedagain partner_identification supports the addition of code to validate a checksum, so it should work.

gurneyalex:
@alexis-via https://github.com/akretion/l10n-france/pull/1 to get a clean travis build. 

and rebase to get a runbot build.

PR #91, opened 2017-01-18
[10.0] Port DES and DEB

alexis-via:
DES has been ported.
Still need to port DEB.


l10n_fr_intrastat_service/data/intrastat_service_reminder.xml
```diff
@@ -0,0 +1,65 @@
...
+<!--
+  © 2013-2017 Akretion (http://www.akretion.com/)
+  @author Alexis de Lattre <alexis.delattre@akretion.com>
+  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+-->
+
+<odoo>
+
+<data noupdate="1"> <!-- noupdate = 1 for the 'active' field -->
+<record id="intrastat_service_cron" model="ir.cron">
```
bealdav, line 11: just a question but in this case, could it works like that ?
```
<data>
<record id="intrastat_service_cron" model="ir.cron">
  <field name="name">Intrastat Service Reminder</field>
  ...
  <field name="args" eval="'()'"/>
</record>
</data>

<data noupdate="1">
<record id="intrastat_service_cron" model="ir.cron">
  <field name="active" eval="False"/>
</record>
</data>
```

alexis-via: I don't think so, because the "noupdate" field is a boolean of ir_model_data, so it can only have 1 value per xmlid.

l10n_fr_intrastat_service/models/intrastat_service.py
```diff
@@ -0,0 +1,378 @@
...
+            return datetime.now().year
+
+    @api.model
+    def _get_month(self):
+        if datetime.now().month == 1:
+            return 12
+        else:
+            return datetime.now().month - 1
+
+    @api.multi
```
bealdav: @api.multi is the default behavior in v10 Model class, then useless, I think

l10n_fr_intrastat_service/models/intrastat_service.py
```diff
@@ -0,0 +1,378 @@
...
+                        amount_invoice_cur_to_write,
+                        self.company_id.currency_id),
+                    0))
+            else:
+                amount_company_cur_to_write = int(round(
+                    amount_invoice_cur_to_write, 0))
+
+            if amount_company_cur_to_write:
+                if invoice.type == 'out_refund':
+                    amount_invoice_cur_to_write = - amount_invoice_cur_to_write
```
bealdav: je connaissais pas ça, je croyais que ct : `a -= b`
TIL that

bealdav (commented):
:+1:

l10n_fr_intrastat_service/data/intrastat_service_reminder.xml
bealdav: exact

l10n_fr_intrastat_service/models/intrastat_service.py
alexis-via: I now use amount_invoice_cur_to_write *= -1
which is more readable.

alexis-via:
FYI, DEB is working fine with this PR.

PR #92, opened 2017-01-26
[9.0][CFONB]Add information in unique_import_id to ensure it is really unique

florian-dacosta:
I have a case where, within the same CFONB file I have 2 identical unique_import_id.
To avoid this case, wen can add additional information into the field and avoid the problem.


alexis-via (approved):
Good idea!
I would just use an intermediary variable "name_ext" to store line[48:79].strip(), to underline that we use twice the same thing.

florian-dacosta:
Done!
Travis error seems unrelated to this PR

PR #95, opened 2017-02-15
[8.0] LCR: FIX don't crash when bank account is missing on a payment line

alexis-via:
It will display an explicit error msg instead of a crash popup.


cdubuit:
👍 thanks

PR #98, opened 2017-03-06
[8.0] mis_report: Add 649 (CICE) account in Profil & Loss

alexis-via:


alexis-via:
I also added the 649000 account in the PCG in this PR on l10n_france : https://github.com/odoo/odoo/pull/15717

Issue #90, opened 2017-01-17
[8.0] l10n_fr_account_fiscal_position_rule crash on install


alexis-via:
On Odoo 8, when trying to install the module l10n_fr_account_fiscal_position_rule, you get this error:
```
Exception: Module loading l10n_fr_account_fiscal_position_rule failed: file l10n_fr_account_fiscal_position_rule/settings/account.fiscal.position.rule.template.csv could not be processed:
 Line 1 : null value in column "fiscal_position_id" violates not-null constraint
DETAIL:  Failing row contains (1, 2017-01-17 17:05:40.799667, 10, null, 1, 1, f, null, null, t, 76, null, null, t, 76, 2017-01-17 17:05:40.799667, null, null, France, both, t, null).
```

This is because, in the module *account_fiscal_position_rule*, the column fiscal_position_id is now required, but, in the module *l10n_fr_account_fiscal_position_rule*, the 29 first rules don't have any fiscal position (and it is the right value in this case).

@sebastienbeau Could you fix this ? It breaks travis on l10n-france/8.0...

alexis-via:
@sebastienbeau What do you want to do with your module l10n_fr_account_fiscal_position_rule ? It breaks Travis in v8...

PR #104, opened 2017-05-05
Set l10n_fr_account_fiscal_position_rule as uninstallable until module is fixed

alexis-via:
Related to bug #90


alexis-via:
@sebastienbeau told me he is OK to disable this module.

PR #106, opened 2017-05-23
[10.0] Add hr_holidays_jours_ouvrables

guewen:
This module allows to handle French leaves based on "Jours Ouvrables".
Not knowing how to translate that, I kept the French words. If you have a better proposal...

### DIFF
--- a/account_banking_fr_lcr/models/account_payment_order.py
+++ b/account_banking_fr_lcr/models/account_payment_order.py
@@ -31,7 +31,9 @@ class AccountPaymentOrder(models.Model):
                 ]
             for unallowed_ascii_char in unallowed_ascii_chars:
                 value = value.replace(unallowed_ascii_char, '-')
-        except:
+        except Exception:
+            # seems that unidecode doesn't raise exception so might
+            # be useless
             raise UserError(_(
                 "Cannot convert the field '%s' to ASCII") % field_name)
         value = value.upper()
--- a/hr_holidays_jours_ouvrables/README.rst
+++ b/hr_holidays_jours_ouvrables/README.rst
@@ -1,0 +1,84 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================
+France - Jours Ouvrables
+========================
+
+Ce module permets de gérer le décompte des jours ouvrables en prenant en
+compte les samedis payés. Lors de la saisie de vacances, le samedi sera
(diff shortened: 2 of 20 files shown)


sebalix:
Maybe "working days"?
We already started a  similar module for an old 6.0 project and the code was based on https://pypi.python.org/pypi/workdays/ so I think this is the right term.

guewen:
> Maybe "working days"?

Isn't it equivocal with the "Jours ouvrés"? If you tell me "working days", I would think about "jours ouvrés".

sebalix:
Yes you are right, I mixed them up.

gurneyalex (approved):

gurneyalex:
@guewen needs rebase

guewen:
Rebase done, conflict resolved (in oca_dependencies.txt).

gurneyalex:
@guewen I rebased again, but travis was red...

account_banking_fr_lcr/models/account_payment_order.py
```diff
@@ -31,7 +31,9 @@ def _prepare_field(self, field_name, field_value, size):
                 ]
             for unallowed_ascii_char in unallowed_ascii_chars:
                 value = value.replace(unallowed_ascii_char, '-')
-        except:
+        except BaseException:
```
sbidoul: In my understanding, in most case it should be `Exception` instead of `BaseException`, because `BaseException` also includes system exiting exceptions.

guewen: @sbidoul that's what I thought too but @yvaucher showed me that `python -m autopep8 --select=E722 -a test.py` replace `except:` by `except BaseException:`
And `except Exception` still raises E722

guewen: I read something wrong without checking :) `except Exception` does not raise E722 hopefully, I fixed that.

yvaucher: It appears we cheched with @guewen that E722 isn't raised with `except Exception:` so it the solution to go.

I think that autopep8 does it to make it exactly the same as `except:` but it is a bad idea in most case to keep it that way for the reason @sbidoul explained.

gurneyalex:
Hello @guewen I tested this on runbot and it looks good to me. Just 1 remark:

* to set the company on the leave type, you need to be in the multicompany group. Maybe this needs to be noted in the readme (non blocking)

Also if others want to test this, remember to set the work address of the employee to France (or the same country where you define public holidays, otherwise they don't get counted.

gurneyalex (approved):

cvinh:
Hello
It appears that this module has not been migrated to version 11.0..14.0... do you know if it has moved to a another repository ?
If not I will start the migration
Thanks in advance

PR #124, opened 2018-01-04
[FIX] broken bank account check

gurneyalex:
when importing bank statements from my customer, I got error about bank
account mismatched caused by the CFONB file including the local account
number, not the full IBAN => improve the check by only testing the relevant
digits.

### DIFF
--- a/account_bank_statement_import_fr_cfonb/models/account_bank_statement_import.py
+++ b/account_bank_statement_import_fr_cfonb/models/account_bank_statement_import.py
@@ -163,3 +163,15 @@ class AccountBankStatementImport(models.TransientModel):
             'transactions': transactions,
             }
         return currency_code, account_number, [vals_bank_statement]
+
+    def _check_journal_bank_account(self, journal, account_number):
+        res = super(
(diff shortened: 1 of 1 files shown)


alexis-via:
Yes, it would be a good remplacement to the code of the usability module that I use to handle this: https://github.com/akretion/odoo-usability/blob/10.0/account_bank_statement_import_usability/account_bank_statement_import.py#L12

The problem is not specific to CFONB files ; it is the same for OFX files in France.

alexis-via (approved):

njeudy:
@alexis-via @gurneyalex it sound like the same problem here ?

https://github.com/OCA/bank-statement-import/pull/136

Will try to see why travis hangs but I think it's a generic way for this problem

pedrobaeza:
@legalsylvain the bot didn't merge it. Do you launch it again?

legalsylvain:
e/ocabot merge patch

legalsylvain:
@gurneyalex could you rebase ? something is wrong on that PR. 

kind regards.

PR #127, opened 2018-01-12
[MIG][11] l10n_siret

fmdl:


legalsylvain (changes requested):
Hi @fmdl. 

Thanks for porting this module ! 
Could you follow the OCA guideline regarding the port of modules ? 
see here : https://github.com/OCA/maintainer-tools/wiki#migration
The process you used delete the commit history and make the review harder to do.
Thanks !

PR #129, opened 2018-01-13
11.0 mig l10n fr state

njeudy:
Migrate to V11 and python 3

PR #131, opened 2018-01-13
11.0 mig l10n fr department

njeudy:
Depends on:

- [X] l10n_fr_state: #129


legalsylvain (changes requested):
Hi @njeudy thanks for porting this module. Some remarks : 
blocking point : why did you removed action and menu item, to manage departments ?
Non blocking point : Could you change yml deprecated file into xml file. 

kind regards.

njeudy:
@legalsylvain hum I will check this tomorrow ..

which yml file are deprecated ?

njeudy:
@legalsylvain I just migrate V10 version of module to V11 I did not remove anything from V10 .. can you explain me which  action / menu item I drop ?

njeudy:
@legalsylvain any news about this ?

legalsylvain:
about yaml files I thought that it was deprecated, but I don't found anything that say such assertion. In other hand, all the OCA and the odoo conventions and guidelines only talk about xml files now.
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#directories

@pedrobaeza, do you have a PoV about that point ?

About action and menu, my mistake, your commit says that you deleted 2 files : 
https://github.com/OCA/l10n-france/pull/131/commits/8a75ffcd547c9662236e84e351093b496b196fe2#diff-3ddce4a9baae483f12e6cd360b266778
but in fact, there two files was deleted before. I don't understand.

regards.

pedrobaeza:
Well, they are deprecated because Odoo has removed the YML import engine in v12, so you will have to switch on next version if you don't do it now.

njeudy:
@pedrobaeza @legalsylvain if it is ok for you I prefer to remove yml file in an other PR. I have a project un V11, and some other module need to be migrated. I will do it in a second stage if it's ok for you ?

@legalsylvain I think we can add an issue if we need action and menu but for this time I just migrate V10 to V11 as is ?

njeudy:
@legalsylvain @pedrobaeza Ok done and rebase to lastest 11.0 branch.

alexis-via (approved):
LGTM. Still a few u'' to remove in manifest. And some "openerp" should be replaced by "odoo".

njeudy:
@alexis-via thanks code update with your comment :)

njeudy:
@legalsylvain ok now for your change request ?

legalsylvain:
Hi. 
I checked again. Understand better. the view and the action still exists, but the menu not.
The problem I see is that the module introduces a model without any menu to manage it. Even if it's stable datas, it is not a good habit if a user want to change a data for any reason.

the menu was existing in 7.0 / 8.0 but was removed here. 
https://github.com/njeudy/l10n-france/commit/34d40c8afee25f9dc4b5a8c38a46673f09da0acd#diff-b20b28aade8e287e795f524a66918eb9R64

could you just add it again ? (in Settings / Technical / Parameters for exemple).

Thanks ! Otherwise LGTM.

alexis-via:
@legalsylvain The menu was commented in v10 because we didn't want to depend on the sales_team module, cf this comment https://github.com/OCA/l10n-france/blob/10.0/l10n_fr_department/view/res_country_department.xml#L64
in v11, menu_localisation belongs to the new "contacts" modules, but we would probably not want to depend on this module either...
But OK to put a menu into Settings / Technical / Parameters... it's not perfect, but it's not our fault... :)

njeudy:
@legalsylvain if someone need a menu, just activate technical mode, and go to the model :) and add menu where your customer need it  ?

We can add process in README ?

legalsylvain:
I don't see what is the problem to add the dropped menu item ? I think it's more work for all to add a comment in the readme and force users to create a menu item.
This module create a model, it should create the according menu item to manage it.

We should not reproduce errors made by odoo. (for res.country for exemple, where the model is defined in base, but the menu item in a sale_xxx module)

alexis-via:
I agree with @legalsylvain

njeudy:
Ok adding menu to contact app ? @alexis-via @legalsylvain

alexis-via:
That means adding a dependancy on the "contacts" module.  But given the content of the contacts module and the fact that it only depends on base, I'm OK for it.

njeudy:
@alexis-via @legalsylvain I add menu, and rebase now that l10n_fr_state is merged.

njeudy:
@alexis-via @legalsylvain all green, ready to merge !! :)

legalsylvain (approved):

PR #137, opened 2018-04-23
[10.0] FIX EBP CSV: wrong order for the 2 last columns

alexis-via:


alexis-via:
Round floats at 2 digits

bealdav (approved):

PR #145, opened 2018-07-05
[FIX] Allow to use FEC report without Siret module

StefanRijnhart:
given that there is no dependency on the siret module in the first place.
Includes some unrelated fixes in the ecotaxe module to make Travis happy.


legalsylvain (approved):
Thanks for this fix and for Making Travis Green Again.

fast tracking this fix.

StefanRijnhart:
Thank you @legalsylvain!

PR #148, opened 2018-07-27
[10.0][FIX] Allow to generate LCR file with res.partner.bank not linked to res.bank

alexis-via:


kryskool (approved):
It's Ok
Works for me

PR #154, opened 2018-10-03
[10.0] l10n_fr_account_tax_unece: post_install now match on XMLID instead of description

alexis-via:
No more duplication of information between post_install and XML data


Doscaal:
:+1:  It's ok. After reinstall this module, tax are configure even the name has been changed

PR #163, opened 2018-12-03
Added module to lookup SIRENE database

remi-filament:
A module created by Le Filament for its French customers is proposed to be integrated in OCA l10n-france repo.
This module adds a wizard on Company partner forms in order to lookup company information in SIRENE database and auto-fills partner legal information accordingly.

This is the first module we push to OCA, so it may very well be that we are not fully compliant with standards. Do not hesitate to let us know if this is the case.


l10n_fr_siren_lookup/.gitignore
```diff
@@ -0,0 +1,3 @@
+.*
```
legalsylvain: Usually OCA repositories doesn't have custom gitignore in the module.

l10n_fr_siren_lookup/LICENSE
```diff
@@ -0,0 +1,230 @@
+GNU AFFERO GENERAL PUBLIC LICENSE
```
legalsylvain: licence is uselesss, as it is pointed in the manifest file.

l10n_fr_siren_lookup/README.rst
```diff
@@ -0,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+
+=======================================================
+Retrieve partner information from French SIREN database
+=======================================================
```
legalsylvain: not a blocking point, but you can now use the new readme templates. 

when PR will be merged, the README.rst file will be generated, based on each readme parts.

see : https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

l10n_fr_siren_lookup/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+
+{
+	'name': 'SIREN Lookup',
+	'summary': "Lookup partner company in French SIREN database",
+	'version': '10.0.1.0',
+	'license': 'AGPL-3',
+	'author': 'LE FILAMENT',
+	'category': 'Partner',
+	'depends': ['base'],
+	'contributors': [
```
legalsylvain: contributors doesn't seem to be a valid key for OCA modules. but you can add a maintainers key, with your github logins.

ref : https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py

l10n_fr_siren_lookup/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+                'Benjamin Rivier <benjamin-filament>',
+		'Rémi Cazenave <remi-filament>',
+        ],
+	'website': 'http://www.le-filament.com',
+	'data': [
+		'wizard/wizard_siren.xml',
+		'views/res_partner.xml',
+	],
+	'qweb': [
+                'static/src/xml/*.xml',
```
legalsylvain: I think that * call is not accepted in OCA modules. could you name each file ?

l10n_fr_siren_lookup/models/res_partner.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+
+# © 2018 Le Filament (<http://www.le-filament.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class PartnerSiren(models.Model):
```
legalsylvain: The name should be ResPartner in that case. (the model name CamelCased)

l10n_fr_siren_lookup/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+
+# © 2017 Le Filament (<http://www.le-filament.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+	'name': 'SIREN Lookup',
+	'summary': "Lookup partner company in French SIREN database",
+	'version': '10.0.1.0',
+	'license': 'AGPL-3',
+	'author': 'LE FILAMENT',
```
legalsylvain: author should include "Odoo Community Association (OCA)"

l10n_fr_siren_lookup/models/res_partner.py
```diff
@@ -0,0 +1,23 @@
...
+	_inherit = 'res.partner'
+
+	## Fields
+	forme_juridique = fields.Char("Legal Type")
```
legalsylvain: French is not allowed in python and JS file. (only in po files).

(same for CHAMPS, etc...)

l10n_fr_siren_lookup/wizard/wizard_siren.py
```diff
@@ -0,0 +1,129 @@
+# -*- coding: utf-8 -*-
```
legalsylvain: generaly there is on file per model. here siren_wizard.py and siren_wizard_company.py

l10n_fr_siren_lookup/wizard/wizard_siren.py
```diff
@@ -0,0 +1,129 @@
...
+	def _default_name(self):
+		return self.env['res.partner'].browse(self.env.context.get('active_id')).name
+
+	@api.model
+	def _default_partner(self):
+		return self.env.context.get('active_id')
+
+	## Fields
+	name = fields.Char(string='Company', default=_default_name)
+	company_lines = fields.One2many('siren.wizard.company', 'wizard_id', string="Results",)
```
legalsylvain: to be consistent with odoo in that case, the name of the field should be ``line_ids`` and the name of the ``comodel siren.wizard.line``

l10n_fr_siren_lookup/wizard/wizard_siren.py
```diff
@@ -0,0 +1,129 @@
...
+		companies = r.json()
+		# Unlink all company lines
+		self.company_lines.unlink()
+		# Fill new company lines
+		for company in companies['records']:
+			new_company = self.company_lines.create({
+					'wizard_id': self.id,
+					'name': company['fields']['l1_normalisee'],
+				})
+			if company['fields'].get('l4_normalisee'):
```
legalsylvain: to be more concise, we use to call _prepare function. 

in that case : 

```

@api.model
def _prepare_partner_from_data(self, data):
    return {
        'street': data.get('l4_normalisee', False),
        'zip': data.get('codpos', False),
    }

```

- It is lighter
- it can be easely overloaded.

l10n_fr_siren_lookup/wizard/wizard_siren.py
```diff
@@ -0,0 +1,129 @@
...
+			'lib_ape': self.lib_ape,
+			'date_creation': self.date_creation,
+			'effectif': self.effectif,
+			'lib_ess': self.lib_ess,
+			'date_ess': self.date_ess,
+			'ess': self.ess,
+			'categorie': self.categorie,
+		})
+
+		return True
```
legalsylvain: I think ``return True`` is useless in V10+

l10n_fr_siren_lookup/wizard/wizard_siren.py
```diff
@@ -0,0 +1,129 @@
...
+# © 2018 Le Filament (<http://www.le-filament.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import json
+import urllib2
+import requests
+
+from odoo import api, fields, models
+
+URL = "https://data.opendatasoft.com/api/records/1.0/search/?dataset=sirene%40public&q="
```
legalsylvain: maybe better

```
URL = "https://data.opendatasoft.com/api/records/1.0/search/?dataset=sirene%40public&q={request}&rows=100"
```

and using later .format(request=req)

legalsylvain (changes requested):
Hi @remi-filament, first : thanks for sharing your code in this OCA.

- Some comments in line.
- in all part, ess should be replaced by sse (social and solidarity economy) or other english words 
- Travis could help you too ref : https://travis-ci.org/OCA/l10n-france/builds/462895909?utm_source=github_status&utm_medium=notification

In a more general point of view, for the generic fields, I see this module good designed (getting, street, zip name, etc...), but for the specific fields (lib_ess, date_ess, etc...) I don't think so. You selected some fields from a big list. https://data.opendatasoft.com/api/records/1.0/search/?dataset=sirene%40public&q=7900585720

So it appears to fit to a custom needs don't you think ?

One of the proposal I could see is to split this module in two modules, a generic one that could be included in the OCA, a custom one that don't. The custom one will be very easy : 
-> overload res.partner and add fields
-> overload prepare functions

@sebastienbeau : it is about json data, do you have a point of view, how to handle it ?

regards.

Note : the data doesn't seems to be good, GRAP is noted as "L’entreprise n’appartient pas au champ de l’ESS" ;-(

l10n_fr_siren_lookup/README.rst
remi-filament: I am not sure how to do that, I suppose that I could have a look once all the other points are resolved :)

remi-filament:
Thank you @legalsylvain , I have solved (I hope) most of your remarks, removed SSE/ESS part from this module to make it more generic
Regarding your Note, this should be up-to-date since database used by INSEE and legal/accounting companies afaik. Although it indeed seems strange that a SCIC would not be tagged as part of SSE (and there is a corresponding SSE date...).

l10n_fr_siren_lookup/wizard/siren_wizard.py
```diff
@@ -0,0 +1,64 @@
...
+    def get_lines(self):
+        # Get request
+        r = requests.get(URL.format(request=self.name))
+        # Serialization request to JSON
+        companies = r.json()
+        # Unlink all company lines
+        self.line_ids.unlink()
+        # Fill new company lines
+        for company in companies['records']:
+            new_company = self.line_ids.create({
```
apollet: I really think you don't need to create records here.
Just put all dict of values in a list and set the line_ids -> self.line_ids = [(6, False, list_of_dict)]
And you'll don't need anymore to do the unlink() before (because nothing is created in database)

remi-filament: Thank you @apollet I have been trying to do what you suggest but I cannot get it to work !

I tried : 
```
        r = requests.get(URL.format(request=self.name))
        lines = []
        for line in r.json()['records']:
            lines.append(line['fields'])
        self.line_ids = [(6, False, lines)]
```

and 
```
        r = requests.get(URL.format(request=self.name))
        self.line_ids = [(6, False, r.json()['records'])]
```

I get the same TypeError: TypeError("unhashable type: 'dict'",)
I think I am missing something here... Any idea ?

legalsylvain: > I really think you don't need to create records here.

I don't understand why. creating a transient record onclicking on a button is a valid way. don't you think ?

Possible improvement : 
```
res = self._prepare_partner_from_data(company['fields'])
res['wizard_id'] = self.id
self.line_ids.create(res)
```

-> you replace a create + write by a single create. (faster)

+ moving ``'name': company['fields']['l1_normalisee']`` in the prepare.

legalsylvain (approved):
Hi @remi-filament thanks for all the changes ! 
LGTM. kind regards.

legalsylvain:
@alexis-via, could you take a look on that PR ?

regards.

l10n_fr_siren_lookup/wizard/siren_wizard.py
apollet: Sorry I was wrong, you have to use the "(0, 0,  { values })" thing.
Try this:

```
def get_lines(self):
    # Get request
    r = requests.get(URL.format(request=self.name))
    # Serialization request to JSON
    companies = r.json()
    # Fill new company lines
    companies_vals = []
    for company in companies['records']:
        res = self._prepare_partner_from_data(company['fields'])
        companies_vals.append((0, 0, res))
    self.line_ids = companies_vals
    return {"type": "ir.actions.do_nothing", } 
```

@legalsylvain Why create records (use database, slowdown user experience) when you don't have to ? :)

legalsylvain: AFAIK, ``self.line_ids = companies_vals`` will call write function of wizard and then create function of wizard line

```
(0, _, values)
    adds a new record created from the provided value dict.
```

Réf : https://www.odoo.com/documentation/12.0/reference/orm.html

alexis-via:
I was initaly a bit reluctant to investigate this PR because I didn't understood why this API from opendatasoft was free (I thought it could become a private API in the future) ; but now that I have a better understanding of the business model of OpenDataSoft, I think that this API will stay open and free.

This API and this initiative to develop an OCA module to use it is really great. Thanks @remi-filament for that !

I have quite a lot of remarks on this module. I implemented some of my remarks in a backport to v8 I made today (cf https://github.com/akretion/l10n-france/commit/449a8c9a7eb8343d8cfd0eeb6f3b376b17ca102d), but I have a lot more remarks (my backport was just a quick backport for one of my users who wanted the feature, not the full changes I would like to make).

Here are my main ideas:
1) we should certainly not define a 'siret' field in this module -> we should depend on l10n_fr_siret
Same remark for APE -> l10n_fr_naf_ape. Otherwise, if both modules are installed, we have 2 fields for the same info (or worst, 2 fields with the same name on the same object in 2 different OCA modules !)
2) As using this API is very simple (no auth) and only requires the "requests" lib, my idea would be to implement the feature directly in l10n_fr_siret, not in a separate module. Then we could have a glue module for l10n_fr_naf_ape if we want. Also, I think we should not add additional fields on res.partner (staff, category, ...) ; these fields would be added by other additional modules (and we use _prepare_ methods to ease the work for those modules)
3) there is a crash when the company name has an accent -> fixed in my backport
4) I implemented a 2 steps wizard (and removed the confirm= on the select button) -> the second step shows the exact information that will be changed on res.partner
5) we should have an easy way to create a partner from it's SIREN or SIRET ; the current implementation only help and getting the SIRET of an existing partner (and updating the address) -> this feature could be used by the module account_invoice_import to automate the creation of the partner is its doesn't already exists.

If you agree with my ideas, I could propose a new implementation in the module l10n_fr_siret that we could merge in OCA.

remi-filament:
Hi @alexis-via,
thank you for taking a look at it.

I agree that there is still some work to do on that one, and my ideas of things to improve match with the ones you listed : 
- Listing the fields that would be updated
- Reuse of existing fields instead of creating new ones

Regarding l10n_fr_siret, I have never used it because from my point of view it is not adequate for our customer needs, let me explain :
1. siret is already present in Odoo since V11.0 in l10n_fr for both res.company and res.partner (it was before only available for res.company - and this module is not creating another one, but reusing the same field but displaying it twice today, this is true).
2. l10n_fr_siret considers that Users have both siren and NIC for their partners, but we see the reality differently : users never have the NIC for their partners, they have more frequently the siret (and sometimes only siren). Therefore, we better should calculate siren from siret and not the other way around.

I think therefore that it is not a good idea to have other modules depending on l10n_fr_siret (see also PR #184 proposing to remove dependency), and even less to implement this feature in l10n_fr_siret module. Please let me know your view ?

alexis-via:
I don't understand your critics against l10n_fr_siret. Your point about the fact that users don't have the NIC but either have SIREN or SIRET is true, but it is exactly the same information at the end. So your point against l10n_fr_siret is about usability, right ? We could certainly improve the usability of l10n_fr_siret. But dropping the module l10n_fr_siret is not the solution ; we should just improve it. l10n_fr_siret adds checksum verification on SIREN/SIRET, which is very important, and it is the only module to provide this feature so far.

remi-filament:
You are right, the checksum verification can be useful to confirm that SIRET is correct.
I still find it counter-intuitive to request the user to enter SIREN + NIC when most of the time he/she does not even know what NIC is (I did not know myself before looking at that module).
From my point of view, to improve usability l10n_fr_siret should ask user to enter directly the SIRET, calculate SIREN from there (whether SIREN is really needed or not is another question though) and verify checksum for correctness.

remi-filament:
Hi @alexis-via, I am trying to clean up a bit mt old PRs.
What should we do with that one ?
We are not working anymore on v10 version so I suppose we could close it and look forward for new implementation on newer Odoo version ?

PR #172, opened 2019-01-16
 [MIG] l10n_fr_chorus_account: Migration to 12.0

joshuajan:


alexis-via:
Replaced by the PR https://github.com/OCA/l10n-france/pull/229 which supports PISTE, the new entry point of the API of AIFE, which will replace the old direct entry point based on RGS certificates.

PR #176, opened 2019-02-21
[WIP] [FIX] Prevent use variable before assignment

vrenaville:


vrenaville:
@alexis-via Could you please check this issue ?

alexis-via:
@vrenaville These variables should be initialized before the FOR loop, but inside it.

PR #181, opened 2019-04-12
[12.0] Migrate DES from v10 to v12

alexis-via:


l10n_fr_intrastat_service/models/intrastat_service.py
```diff
@@ -0,0 +1,378 @@
...
+        # delete all DES lines generated from invoices
+        lines_to_remove = line_obj.search([
+            ('invoice_id', '!=', False),
+            ('parent_id', '=', self.id)])
+        lines_to_remove.unlink()
+
+        invoices = invoice_obj.search(
+            self._prepare_domain(), order='date_invoice')
+        for invoice in invoices:
+            if not invoice.partner_id.country_id:
```
ThomasBinsfeld: Shouldn't it be based on `commercial_partner_id`?

l10n_fr_intrastat_service/models/intrastat_service.py
```diff
@@ -0,0 +1,378 @@
...
+                line_obj.create({
+                    'parent_id': self.id,
+                    'invoice_id': invoice.id,
+                    'partner_vat': partner_vat_to_write,
+                    'partner_id': invoice.partner_id.id,
+                    'invoice_currency_id': invoice.currency_id.id,
+                    'amount_invoice_currency': amount_invoice_cur_to_write,
+                    'amount_company_currency': amount_company_cur_to_write,
+                    })
+        self.message_post(_('Re-generating lines from invoices'))
```
ThomasBinsfeld: This line produces an error: `TypeError: message_post() takes 1 positional argument but 2 were given`.

```suggestion
        self.message_post(body=_('Re-generating lines from invoices'))
```

ThomasBinsfeld (changes requested):

l10n_fr_intrastat_service/models/intrastat_service.py
alexis-via: I also get this crash. But, when I look at the proto of the method, it seems ok : https://github.com/odoo/odoo/blob/12.0/addons/mail/models/mail_thread.py#L1931
The named arg "body" is the first argument of message_post(), so it should work without "body=", no ?

alexis-via: message_post() had a similar proto on odoo v10, so maybe it is a change of behavior between python2 and python3...

l10n_fr_intrastat_service/models/intrastat_service.py
ThomasBinsfeld: ping @alexis-via

alexis-via: yes, I'm not against changing that...
But I would really like to understand the problem with message_post()... it seems to me it is a different between python 2 and 3, but I'm still not sure about it.

l10n_fr_intrastat_service/models/intrastat_service.py
sbidoul: @alexis-via the problem is that there are all kinds of message_post overrides, such as `def message_post(self, **kwargs):`. So this method is really intended to be used with keyword arguments only.

Actually in python3, it should better be declared like this: `def message_post(self, *, body='', ...)` to force users to call its with keyword arguments.

sbidoul: https://github.com/odoo/odoo/pull/33306

PR #184, opened 2019-06-04
[IMP][l10n_fr_intrastat_product] Remove l10n_fr_siret dependency

ivantodorovich:
It's actually not needed. We already have `company_id.siret` in `l10n_fr`.
l10n_fr_siret adds siret computation based on siren and nic. It's nice, but not needed in this module.

### DIFF
--- a/l10n_fr_intrastat_product/__manifest__.py
+++ b/l10n_fr_intrastat_product/__manifest__.py
@@ -4,15 +4,15 @@
 
 {
     'name': 'DEB',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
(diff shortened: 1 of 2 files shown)


ivantodorovich:
btw travis is failing because:

> odoo.exceptions.UserError: ("You try to install module 'account_balance_ebp_csv_export' that depends on module 'account_financial_report'.\nBut the latter module is not available in your system.", '')

account_balance_ebp_csv_export is on this repo, but account_financial_report is not migrated yet

ivantodorovich:
ping @alexis-via 
can you check?

alexis-via (approved):

PR #192, opened 2019-10-08
[FIX] Bug fix for XML declaration generation

vrenaville:
* Purpose
Without this fix it's not possible to install the module, and the generation of xml file failed (without the python fix)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ cache:
     - $HOME/.cache/pip
 
 addons:
+  postgresql: "9.6"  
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -9,6 +9,7 @@ cache:
     - $HOME/.cache/pip
 
 addons:
+  postgresql: "9.6"
```
Tonow-c2c, line 12: not upper ?

gurneyalex (approved):

PR #194, opened 2019-10-24
[12.0] [ADD] l10n_fr_cedex module

remi-filament:
Cedex field added on res_partner

### DIFF
--- a/l10n_fr_cedex/__init__.py
+++ b/l10n_fr_cedex/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/l10n_fr_cedex/__init__.py
+++ b/l10n_fr_cedex/__init__.py
@@ -1,1 +1,3 @@
 from . import models
+from .hooks import post_init_hook
+from .hooks import uninstall_hook
--- a/l10n_fr_cedex/__manifest__.py
+++ b/l10n_fr_cedex/__manifest__.py
@@ -1,0 +1,17 @@
+# © 2019 Le Filament (<https://le-filament.com>)
+
+{
+    'name': 'Add Cedex on Partner',
+    'version': '12.0.1.0.0',
+    'category': 'Extra Tools',
+    'license': 'AGPL-3',
+    'summary': 'Adds Cedex on Partner',
+    'author': "Le Filament,Odoo Community Association (OCA)",
+    'website': 'https://le-filament.com',
+    'depends': ['base'],
+    'installable': True,
+    "data": [
+        "datas/res.country.csv",
+        "views/res_partner.xml",
+    ]
+}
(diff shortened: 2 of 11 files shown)


flotho (approved):
LGTM

legalsylvain (changes requested):
Thanks for this module @remi-filament.

1) Testing on runbot, to check if cedex field is displayed on report, it is generating a bug.
Step to reproduce : 
- set cedex on a partner
- try to print a quotation for this partner. (for instance, purchase.order).

Server error : 

```
  File "/.repo_requirements/odoo/odoo/addons/base/models/res_partner.py", line 695, in _get_name
    name = name + "\n" + partner._display_address(without_company=True)
  File "/.repo_requirements/odoo/odoo/addons/base/models/res_partner.py", line 936, in _display_address
    return address_format % args
KeyError: 'cedex'

Error to render compiling AST
KeyError: 'cedex'
Template: purchase.report_purchasequotation_document
Path: /templates/t/t/t[4]/t/div/div
Node: <div t-field="o.picking_type_id.warehouse_id.partner_id" t-options="{&quot;widget&quot;: &quot;contact&quot;, &quot;fields&quot;: [&quot;address&quot;, &quot;phone&quot;], &quot;no_marker&quot;: True, &quot;phone_icons&quot;: True}"/>
              
```

2) I also think that the field should be available on ``res.company`` form that inherit from ``res.partner``.

remi-filament:
Thanks for the review @legalsylvain, you are right, I forgot half of the module...
I added cedex field to _address_fields() for proper render.
I realized that the \n from csv file were not taken correctly into account, so I replaced it by hooks for updating database directly.
If you could check again, it should be working now !

remi-filament:
And porting on res_company as requested

legalsylvain (commented):
Hi @remi-filament 

Much better ! 

Tested on runbot.
- set cedex on res.company : OK
- display cedex on report for res.company : OK

- display cedex for res.partner : KO. (or maybe I missed something.)

See below, i set CEDEX on the current company, and the current customer. 

image

question : don't we have to add a constraint on the new field ? 
cedex field should be formated like the following string :  

CEDEX XX 
where XX is a number. 

don't you think ? 

ref : https://fr.wikipedia.org/wiki/Courrier_d%27entreprise_%C3%A0_distribution_exceptionnelle

remi-filament:
@legalsylvain To display partner cedex on report, country needs to be France, any DOM-TOM, Andorra, or Monaco. Otherwise it makes no sense from my point of view.

Regarding the constraint, I am not sure, since sometimes you can have only CEDEX, sometimes CEDEX X, sometimes CEDEX XX... We could force to have the field starting with CEDEX but I am not sure whether mandatory ?

PR #206, opened 2020-01-17
Allow for CFONB files without linebreaks

tschanzt:


jcoux (changes requested):
Can you add comment in code and explain your modification in PR description?
Also example of required content files can help to understand.

tschanzt:
@jcoux i added a comment

jcoux (approved):
Thanks for explanations.

account_bank_statement_import_fr_cfonb/wizard/account_bank_statement_import.py
```diff
@@ -57,21 +57,21 @@ def _parse_file(self, data_file):
...
                 data_file)
         transactions = []
-        if not data_file.splitlines():
+        # The CFONB spec says you should only have digits, capital letters
+        # and * - . /
+        # But many banks don't respect that and use regular letters for exemple
+        # And I found one (LCL) that even uses caracters with accents
+        # So the best method is probably to decode with latin1
+        data_file_decoded = data_file.decode('latin1')
+        if not self.split_lines(data_file_decoded):
```
i-vyshnevska:
```suggestion
        lines = self.split_lines(data_file_decoded)
        if not lines:
        
```

account_bank_statement_import_fr_cfonb/wizard/account_bank_statement_import.py
```diff
@@ -57,21 +57,21 @@ def _parse_file(self, data_file):
...
         start_balance = end_balance = False
         vals_line = False
-        # The CFONB spec says you should only have digits, capital letters
-        # and * - . /
-        # But many banks don't respect that and use regular letters for exemple
-        # And I found one (LCL) that even uses caracters with accents
-        # So the best method is probably to decode with latin1
-        data_file_decoded = data_file.decode('latin1')
-        for line in data_file_decoded.splitlines():
+        for line in self.split_lines(data_file_decoded):
```
i-vyshnevska:
```suggestion
        for line in lines:
```

account_bank_statement_import_fr_cfonb/wizard/account_bank_statement_import.py
```diff
@@ -158,5 +158,32 @@ def _parse_file(self, data_file):
...
+        data_file_without_linebreaks = data_file.replace(
+            '\n', ''
+        ).replace('\r', '')
+
+        # check length of file
+        max_len = len(data_file_without_linebreaks)
+        lines = []
+
+        # make sure the length is a multiple of 120 otherwise it isn't valid
+        if max_len % 120:
```
i-vyshnevska: 120 can be a constant in this case

account_bank_statement_import_fr_cfonb/wizard/account_bank_statement_import.py
```diff
@@ -57,21 +57,21 @@ def _parse_file(self, data_file):
...
         transactions = []
-        if not data_file.splitlines():
+        # The CFONB spec says you should only have digits, capital letters
+        # and * - . /
+        # But many banks don't respect that and use regular letters for exemple
+        # And I found one (LCL) that even uses caracters with accents
+        # So the best method is probably to decode with latin1
+        data_file_decoded = data_file.decode('latin1')
+        if not self.split_lines(data_file_decoded):
             raise UserError(_('The file is empty.'))
```
i-vyshnevska:
```suggestion
```
can be also raised in split_lines

account_bank_statement_import_fr_cfonb/wizard/account_bank_statement_import.py
sebalix: So we are enforcing the CFONB spec here, with maybe generating regressions in current module installations?
Should we fallback on the previous behavior (so splitting lines whatever the size of the lines are)? It is just a question, not blocking. If every banks respect the rule of 120 characters we are fine.

tschanzt: Well that is what i got from @CharlineDumontet, i would imagine the 120 chars are more important if there is banks sending the file around without newline. Sadly i can not read the spec as my french is not good enough to navigate the cfonb site to find it (i suspect there might not be a english version of the spec available as well)

sebalix (approved):

PR #210, opened 2020-02-13
[10.0] New module for DAS2

alexis-via:
New module for DAS2, also called "Décalaration d'honoraires".

The DAS2 file I sent today to the fiscal administration was accepted as valid.

I still have some PEP8 work and code clean-up to do.


alexis-via:
The code is now clean. It is ready for merge.

PR #209, opened 2020-02-13
FIX travis in l10n-france

bealdav:

PR #212, opened 2020-02-21
[12.0] l10n_fr_fec_oca: restore working code

alexis-via:
Restore working code following this PR https://github.com/OCA/l10n-france/pull/209/files which broke the module


bealdav:
It breaks tests, I'll work on it to pass tests

PR #213, opened 2020-02-21
[10.0] FEC OCA - Backport from v12

alexis-via:


alexis-via:
And there is now a new option "include initial balance", which is not (yet) in v12.

PR #214, opened 2020-02-26
12.0 - FIX prevent install l10n_fr_account_invoice_facturx in travis settings because it break tests

bealdav:
It's patch while we find a better way to test this module without break repository.
Maybe remove `auto_install` of `account_facturx` module in OCB


bealdav:
please @legalsylvain @alexis-via @sebastienbeau 
could you have a look ?

sebastienbeau:
We should it to avoid breaking everything and then later fix it in a different ways

PR #222, opened 2020-04-28
[10.0] cfonb: improve handling of lines type 05

alexis-via:


alexis-via:
See comment in code for more explanations about the problem and the solution.

PR #229, opened 2020-05-21
[12.0] MIG Chorus modules from v10 to v12

alexis-via:
Roadmap: add the OCA solution for pwd


alexis-via:
Requires PR https://github.com/OCA/contract/pull/520

PR #240, opened 2020-06-23
[12.0] siret: FIX readonly status on company fields

alexis-via:


ivantodorovich (approved):

l10n_fr_siret/models/company.py
```diff
@@ -12,11 +12,11 @@ class ResCompany(models.Model):
 
     # siret field is defined in l10n_fr module on res.partner
     # with an unstored related field on res.company
-    siret = fields.Char(store=True)
+    siret = fields.Char(store=True, readonly=True)
     siren = fields.Char(
-        string='SIREN', related='partner_id.siren', store=True)
+        string='SIREN', related='partner_id.siren', store=True, readonly=False)
```
bealdav, line 17: For my information, I thought than write on related fields should be avoided causing perf issues.
Maybe it's a misunderstood on my side ?
Here if you click on partner m2o you can edit these fields with no real inconvenience.

alexis-via: I'm not aware of such write perf issues, but writing on res.company is not something that is done very often, so it's not really a problem.
Without this fix, the SIRET field on res.company was writable, but it wouldn't update the siren/nic field, and at the end the siret field of res.partner was empty. So the aim is to fix this bug, nothing else.

bealdav: ok

kryskool (approved):

pedrobaeza:
@legalsylvain the bot didn't merge it. Do you launch it again?

legalsylvain:
Thanks pedro.

PR #243, opened 2020-07-28
[ADD] hr_check_ssnid module

Tonow-c2c:
image
Peek 2020-07-28 11-31


hr_check_ssnid/models/hr_employee.py
```diff
@@ -0,0 +1,53 @@
...
+    _inherit = 'hr.employee'
+
+    ssnid_key = fields.Char(
+        string='SSN No Key',
+        help='Social Security Number Key',
+        size=2,
+        required=True,
+    )
+
+    @api.onchange('ssnid', 'ssnid_key') # if these fields are changed, call method
```
ivantodorovich: Shouldn't this be an `@api.constraint` ?
You can probably apply it directly in `check_ssnid` method.
Although remember `self` is going to be a recordset, so `for rec in self:`

hr_check_ssnid/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+ 'author': 'Camptocamp,Odoo Community Association (OCA)',
+ 'license': 'AGPL-3',
+ 'category': 'French Localization',
+ 'depends': ['hr',
+             ],
+ 'website': 'https://www.camptocamp.com',
+ 'data': ['views/hr_employee.xml',
+          ],
+ 'installable': True,
+ 'external_dependencies': {'python': ['stdnum']}
```
ivantodorovich: also add `stdnum` to `requirements.txt` file in repository root.
If the file isn't there, create it.

ivantodorovich (changes requested):
As for module name, I think it should be `l10n_fr_hr_ssnid_validation`

hr_check_ssnid/__manifest__.py
Tonow-c2c: in other OCA repo I didn't see `requirement.txt` link https://github.com/OCA/geospatial/blob/12.0/base_geoengine_demo/__manifest__.py#L22

https://github.com/OCA/geospatial

sebalix: @Tonow-c2c but if it exists it'll be read by the Travis configuration (e.g. https://github.com/OCA/report-print-send/)

Tonow-c2c: Thanks :+1:

l10n_fr_hr_ssnid_validation/views/hr_employee.xml
```diff
@@ -0,0 +1,18 @@
...
+<odoo>
+
+  <record id="view_employee_form" model="ir.ui.view">
+    <field name="name">ssnid.view.employee.form</field>
+    <field name="model">hr.employee</field>
+    <field name="inherit_id" ref="hr.view_employee_form"/>
+    <field name="arch" type="xml">
+
+      <field name="user_id" position="after">
+        <field name="ssnid"/>
```
i-vyshnevska: isn't weird that field ssnid defined in core is not shown anywhere in UI in odoo?

l10n_fr_hr_ssnid_validation/models/hr_employee.py
```diff
@@ -0,0 +1,54 @@
...
+    ssnid_key = fields.Char(
+        string='SSN No Key',
+        help='Social Security Number Key',
+        size=2,
+        required=True,
+    )
+
+    @api.onchange('ssnid', 'ssnid_key')
+    def onchange_ssnid(self):
+        self.check_ssnid()
```
i-vyshnevska: isn't constrain not enough?

l10n_fr_hr_ssnid_validation/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
```
i-vyshnevska:
```suggestion
# Copyright 2020 Camptocamp SA
```

l10n_fr_hr_ssnid_validation/models/hr_employee.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @api.constrains('ssnid', 'ssnid_key')
+    def check_ssnid(self):
+        # We assume this dev work only in France
+        # other available format are on
+        # https://arthurdejong.org/python-stdnum/doc/1.13/index.html#available-formats
+        mod = get_cc_module('fr', 'nir')
```
i-vyshnevska: can you rely on company employee belongs to? it will prevent errors for multicompany mode

i-vyshnevska (changes requested):

l10n_fr_hr_ssnid_validation/models/hr_employee.py
sebalix: The constraint apply when pushing the Save button/write call I think, and it could help the user if he is warned directly when changing the value on the form?

i-vyshnevska: as it is two fields it will be called when one field empty and of course will fail the check, that is annoying

l10n_fr_hr_ssnid_validation/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{'name': 'France - Ssnid check',
+ 'version': '10.0.1.0.0',
+ 'author': 'Camptocamp,Odoo Community Association (OCA)',
+ 'license': 'AGPL-3',
+ 'category': 'French Localization',
+ 'depends': ['hr',
+             ],
+ 'website': 'https://www.camptocamp.com',
```
legalsylvain: Please set github url.

l10n_fr_hr_ssnid_validation/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
```
legalsylvain: The header are useless in python3.

legalsylvain (changes requested):
Some comment inline.

Otherwise it seems you are changing the meaning of the field ssnid. 
I think you should introduce another field ssnid_nir char(13), and compute ssnid with nir and key.

l10n_fr_hr_ssnid_validation/models/hr_employee.py
```diff
@@ -0,0 +1,54 @@
...
+        size=2,
+        required=True,
+    )
+
+    @api.onchange('ssnid', 'ssnid_key')
+    def onchange_ssnid(self):
+        self.check_ssnid()
+
+    @api.constrains('ssnid', 'ssnid_key')
+    def check_ssnid(self):
```
ivantodorovich: Self is a recordset here.
You need to loop through records otherwise it'll fail when writing on multiple records

l10n_fr_hr_ssnid_validation/__init__.py
Tonow-c2c: it's version 10.0 so python 2

Tonow-c2c:
Sorry a change again the module name to be explicit to the fact I add the `ssnid_key` on this module

legalsylvain:
> Sorry a change again the module name to be explicit to the fact I add the ssnid_key on this module

Sorry, I think the previous module name was correct. But the design not !
Without the module installed, the ``ssnid`` field contain a 15 (13 + 2) char. With the module installed, it seems that the ``ssnid`` field contains a 13 char (in your screen shot, and in the test) So, the module is changing the meaning of a field, that is not attended. Adding two fields ``ssnid_nir`` and ``ssnid_key`` should fix the problem. (and make ``ssnid = ssnid_nir + ssnid_key``
Sorry if I was not clear in my previous comment.

I also don't understand where the following process is done : 
``The original ssnid is split to ssnid + ssnid_key to help if previous data are wrong.``

kind regards.

Regarding the design, I think it's similar to siren / siret. the ``siret = siren + nic`` (with siret defined in Odoo core, and siren & nic defined in extra oca module)
ref : https://github.com/OCA/l10n-france/blob/12.0/l10n_fr_siret/models/partner.py#L68

@alexis-via : could you take a look on this PR ? 

CC : @robinkeunen, @quentinDupont : This module could interest you.

l10n_fr_hr_ssnid_validation/__init__.py
legalsylvain: oh, indeed ! Sorry for the noice.

lmignon:
@Tonow-c2c IMO, this kind of identication numbers should be implemented using https://github.com/OCA/partner-contact/tree/10.0/partner_identification
We did it this way for KBO/BCE numbers into l10n_belgium https://github.com/OCA/l10n-belgium/tree/10.0/l10n_be_partner_kbo_bce

Tonow-c2c:
I close this PR https://github.com/OCA/l10n-france/tree/10.0/l10n_fr_hr_check_ssnid already exist

Tonow-c2c:
I close this PR https://github.com/OCA/l10n-france/tree/10.0/l10n_fr_hr_check_ssnid already exist

alexis-via:
Yes, please port the module l10n_fr_hr_check_ssnid (available on v10) to v12, instead of developping a new module.
And it's very weired to use 2 fields, one for the number without the key and another one for the key. We should just use one field for the full number (key included). That's how it is handled in l10n_fr_hr_check_ssnid.

alexis-via:
I just found out that this PR was for v10... where the module l10n_fr_hr_check_ssnid already exists ! :)

PR #252, opened 2020-11-01
14.0 mig l10n fr siret

rvalyi:
standard migration according to https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0#technical-method-to-
migrate-a-module-from-130-to-140-branch
and fixed translation from v12 (the only diff between v12/v13)
cc @alexis-via @hparfr


rvalyi:
The pre-commit test is currently failing. This was due to the pre-commit itself. Then I tried to apply the patch being tested by @sbidoul but another issue (re-)appeared as I commented in https://github.com/OCA/oca-addons-repo-template/pull/38
Once this oca-addons-repo-template issue is fixed I will force push the PR again with only the expected commits.

rvalyi:
I rebased and the pre-commit issues have been fixed.

hparfr (approved):
LGTM

bealdav (approved):

rvalyi:
I just got aware @alexis-via commented about basing this module on https://github.com/OCA/partner-contact/tree/12.0/partner_identification
So may be worth double checking what we do.

alexis-via:
In the end, I don't think we'll base it on partner_identification.

But I'd like to introduce another (important) change in this module.
As you know, the SIREN is for a whole company. But if the company has multiple adresses, it will have one NIC per address. So, on the parent partner (that should have the address of the headquarters), the NIC should be the NIC of the headquarters. But it should be possible to have a different NIC on some contacts/delivery/invoicing adresses that are different from the address of the headquarters.
That would involve 3 changes :
1) Remove 'nic' from _commercial_fields()
2) Add nic to _address_fields()
3) Add nic in contact form view. As the use_parent_address field is gone (grrrr !), it would be displayed on the invoice address/delivery address/other address.
In the code of modules that use l10n_fr_siret, when they want to get the SIRET of the company like before, they should use partner_id.commercial_partner_id.siret (this is something that I'll check when porting the modules that depend on l10n_fr_siret).
Do you agree with this change ? @sebastienbeau @fclementic2c @fmdl

fmdl:
@alexis-via no comment, I don't use this module, because siret already exist in odoo. (I think this module is not useful)

rvalyi:
Hello @alexis-via I was about to ask you about partner_identification, I had no time yesterday but now that you gave me the reasons let me read that email again and confirm (cause we have the very same doubt in Brazil). Thanks for the track_visibility conversion I got the issue only after Runbot started running.

alexis-via:
@fmdl You say it's not useful because you don't use it!!! This module checks the checksum of SIRET (many people don't know that SIRET have a checksum !), and it avoids a lot of mistakes ! And having mistakes in SIRET leads to problems in Chorus Pro for example.

fmdl:
I have don't that there are a checksum compute (maybe ask Odoo to add in standard). The checksum compute is a good things.

alexis-via:
My PR about the change on NIC is here: https://github.com/akretion/l10n-france/pull/7/files (it points to this branch).

PR #255, opened 2020-11-18
[14.0][MIG] account_statement_import_fr_cfonb

alexis-via:
Depends on https://github.com/OCA/bank-statement-import/pull/359


account_statement_import_fr_cfonb/wizard/account_statement_import.py
```diff
@@ -0,0 +1,206 @@
...
+            elif rec_type == "04":
+                amount = self._parse_cfonb_amount(line[90:104], decimals)
+                ref = line[81:88].strip()  # This is not unique
+                name = line[48:79].strip()
+                transactions.append(
+                    {
+                        "sequence": seq,
+                        "date": date_dt,
+                        "payment_ref": name,
+                        "unique_import_id": "{}-{}-{:.2f}-{}".format(
```
fmdl, line 140: attention il peut avoir deux transactions avec le même `unique_import_id` (cela arrive avec des billets d'avion)

account_statement_import_fr_cfonb/wizard/account_statement_import.py
```diff
@@ -0,0 +1,206 @@
...
+                or guichet_code != line_guichet_code
+                or currency_code != line_currency_code
+                or account_number != line_account_number
+            ):
+                raise UserError(
+                    _("The CFONB file is inconsistent. Error on line %d.") % i
+                )
+        return result
+
+    def split_lines(self, data_file):
```
fmdl: mettre en privé `def _split_lines()`

account_statement_import_fr_cfonb/wizard/account_statement_import.py
```diff
@@ -0,0 +1,206 @@
...
+        if amount_str[-1] in debit_trans:
+            amount_num = float("-" + amount_str[:-1] + debit_trans[amount_str[-1]])
+        elif amount_str[-1] in credit_trans:
+            amount_num = float(amount_str[:-1] + credit_trans[amount_str[-1]])
+        return amount_num
+
+    def _check_journal_bank_account(self, journal, account_number):
+        res = super()._check_journal_bank_account(journal, account_number)
+        if not res:
+            # compare the bank account number only instead of the full IBAN
```
fmdl: ne comparer uniquement si c'est un compte français.
```python
if jrnl_full_acc_number[0:2] != 'FR':
    return res
```

alexis-via: I removed that code. It's now handled in account_statement_import.

account_statement_import_fr_cfonb/wizard/account_statement_import.py
alexis-via: done.

account_statement_import_fr_cfonb/wizard/account_statement_import.py
alexis-via: Yes, it's not new, this problem has always existed. What do you propose ? Remove unique_import_id completely ? That would disable the auto-drop of already imported lines. My idea: we could make it a configuration option on res.company. What do you think ?

fmdl:
```python
def _cfonb_unify_transaction(self, transactions):
        """ If case there two operation in the same day, with same amount
            and same ref. """
        unique_import_ids = {}
        for transaction in transactions:
            unique_import_id = transaction['unique_import_id']
            if unique_import_id in unique_import_ids.keys():
                transaction['unique_import_id'] += '-%s' % unique_import_ids[unique_import_id]
                unique_import_ids[unique_import_id] += 1
            else:
                unique_import_ids[unique_import_id] = 1
        return transactions
```

luc-demeyer (approved):
Tested in combination with https://github.com/Noviat/account_ebics/tree/14.0-mig-account_ebics and https://github.com/OCA/bank-statement-import/pull/359.
A small bug needs to corrected in https://github.com/OCA/bank-statement-import/pull/359 (cf. my requested changes) but this PR seems to be ok.

account_statement_import_fr_cfonb/wizard/account_statement_import.py
alexis-via: OK, I integrated your suggestion, I think it's the least worse solution.

PR #256, opened 2020-11-18
[13.0][FIX] account_banking_fr_lcr: proper rounding when amount is 15.99999999 in DB

alexis-via:


alexis-via:
Approved by the user who reported the bug to me.

PR #258, opened 2020-11-30
12.0 rup

sebastienbeau:
@kevinkhao


l10n_fr_rup/report/report_rup.xml
```diff
@@ -0,0 +1,191 @@
...
+                    </td>
+                </tr>
+            </tbody>
+        </template>
+
+        <template id="report_l10n_fr_rup">
+            <style type="text/css">
+                <t t-call="l10n_fr_rup.style_rup"/>
+            </style>
+            <t t-call="web.html_container">
```
legalsylvain: Do you have the classic header and footer if you inherit of ``web.html_container`` ? (logo of the company, etc...)

l10n_fr_rup/models/hr_contract.py
```diff
@@ -0,0 +1,31 @@
...
+
+from odoo import models, fields, api
+
+
+class HrContract(models.Model):
+    _inherit = "hr.contract"
+    _order = "date_start,name"
+
+    pcs_id = fields.Many2one("hr.employee.pcs", "PCS")
+    employer_address_id = fields.Many2one("res.partner", "Addresse de l'employeur")
```
legalsylvain: address_id and work_location doesn't seems to be required field by law. Or did I missed something ? 
https://www.service-public.fr/professionnels-entreprises/vosdroits/F1784

legalsylvain (commented):
hi @sebastienbeau.
Thanks for this work. Some remarks. Not sure all is relevant though. @quentinDupont : maybe it could interest you.

- Could you add a link and a small text, regarding what is rup ? I had to search in the summary, to understand that it was talking about registre unique du personnel. Maybe in the title, it's better.
- Also I think that the name of the module should be ``l10n_fr_hr_rup`` or l10n_fr_hr_contract_rup``.

- I'm not sure to understand what is PSC ? Catégories sociaux professionnels ? https://fr.wikipedia.org/wiki/PSC ? ;-)
Is it the "emploi" field, described here ? https://code.travail.gouv.fr/fiche-ministere-travail/le-registre-unique-du-personnel


Side question : we should have a register per "établissement secondaire" how do you handle that point ?

l10n_fr_rup/report/report_rup.xml
kevinkhao: No, this report was meant basically as a very simple formatted table.

kevinkhao:
@legalsylvain yes, we changed from CSP (catégorie socioprofessionelle) to PCS. For the other points I think this module will be extended in time

legalsylvain:
> @legalsylvain yes, we changed from CSP (catégorie socioprofessionelle) to PCS. For the other points I think this module will be extended in time

as Professional Category Social ? I think it should be written in full, in a description of the module.

l10n_fr_rup/report/report_rup.xml
legalsylvain: I think adding classical header and legal information present in the footer could be a great addition. (if not in the scope of this PR, could be added in the roadmap section).

sebastienbeau:
@kevinkhao can you process the feedback of Sylvain ? thanks

kevinkhao:
@sebastienbeau @legalsylvain 
- [x] Rename to l10n_fr_hr_rup
- [x] Add readme fragments and write out terms PCS, RUP
- [x] Add roadmap
- [ ] Make a decision on address_id and work_location (probably useful, but should we keep it or remove it ?)

l10n_fr_rup/models/hr_contract.py
kevinkhao: employer_address_id refers to the temporary work contractor ("agence d'intérim") company address.
Official government page doesn't give any example of a RUP. This was made in accordance to our client's RUP, so we presume it is compliant and leave it as it is. If there are other concrete examples we can draw inspiration from maybe we can make some modifications.

kevinkhao:
@legalsylvain 
employer_address_id refers to the temporary work contractor ("agence d'intérim") company address.
Official government page doesn't give any example of a RUP. This was made in accordance to our client's RUP, so we presume it is what is used in reality and leave it as it is. If there are other concrete examples we can draw inspiration from maybe we can make some modifications.

alexis-via:
(remark copied from my remark on the v14 PR)
Looking at the report, I understand that you're storing the social security number in the field identification_id. I think you should store it in the field ssnid and add a dependency on the module l10n_fr_hr_check_ssnid which checks that the social security number has a good checksum.

PR #260, opened 2020-12-01
[14.0][MIG] l10n_fr_fec_oca

alexis-via:


l10n_fr_fec_oca/wizard/account_fr_fec_oca.py
```diff
@@ -0,0 +1,579 @@
...
+
+    @api.model
+    def _default_partner_account_ids(self):
+        pay = self.env['ir.property']._get(
+            'property_account_payable_id', 'res.partner')
+        rec = self.env['ir.property']._get(
+            'property_account_receivable_id', 'res.partner')
+        return pay + rec
+
+    def do_query_unaffected_earnings(self):
```
fmdl, line 106: il serait bien de mettre cette méthode en privée (du fait que c'est du pur SQL) + rajouter un `hasgroup`.

l10n_fr_fec_oca/wizard/account_fr_fec_oca.py
```diff
@@ -0,0 +1,579 @@
...
+                _("Missing VAT number for company %s") % company.name)
+        vat = company.vat.upper().replace(' ', '')
+        if not is_dom_tom and vat[0:2] != 'FR':
+            raise UserError(
+                _("FEC is for French companies only !"))
+
+        siren = vat[4:13] if not is_dom_tom else ''
+        return siren
+
+    def generate_fec(self):
```
fmdl, line 198: idem, il serait bien de rajouter user.hasgroup(account_manager)

l10n_fr_fec_oca/wizard/account_fr_fec_oca.py
alexis-via: Now that we have ir.model.access on wizards in v14, I think it's not usefull to have a hasgroup(), no ? I agree that it would be useful on v13-.

l10n_fr_fec_oca/wizard/account_fr_fec_oca.py
```diff
@@ -0,0 +1,579 @@
...
+        ('accounts', 'Selected Accounts'),
+        ('all', 'All'),
+        ], default='types', required=True, string='Partner Export Option')
+    partner_account_type_ids = fields.Many2many(
+        'account.account.type', string='Account Types',
+        default=lambda self: self._default_partner_account_type_ids())
+    partner_account_ids = fields.Many2many(
+        'account.account', string='Accounts',
+        default=lambda self: self._default_partner_account_ids())
+    fec_data = fields.Binary('FEC File', readonly=True, attachment=False)
```
florian-dacosta: @alexis-via I wonder, why `attachment=False` ?
I believe it should be changed. Indeed, I had an issue in version 9 with this error : 
`File "/odoo/src/openerp/sql_db.py", line 233, in executeres = self._obj.execute(query, params)InternalError: invalid memory alloc request size 1073741824`
When Odoo try to write the binary into the database.
It seems the file was too big for postgresql... Storing in in the filestore resolve the issue. 

Of course, it only happen when we have a lot of data. 
Can we consider changing this ?

alexis-via: I've put attachment=False because it is like that in the native module, cf https://github.com/odoo/odoo/blob/14.0/addons/l10n_fr_fec/wizard/account_fr_fec.py#L20
Feel free to change it if you think it is better.

florian-dacosta: Done !

fmdl:
FYI : https://github.com/odoo/odoo/pull/65170

PR #265, opened 2020-12-21
[MIG] 14.0 hr_rup

PierrickBrun:
based on https://github.com/OCA/l10n-france/pull/258


PierrickBrun:
I think functional reviews will be needed because I am not sure of what should the result look like @kevinkhao

alexis-via:
Looking at the report, I understand that you're storing the social security number in the field identification_id. I think you should store it in the field ssnid and add a dependency on the module l10n_fr_hr_check_ssnid which checks that the social security number has a good checksum.

kevinkhao:
> I think functional reviews will be needed because I am not sure of what should the result look like @kevinkhao

As far as we know the result fits the spec for our client and it looks like the physical RUP that they have been using. As more people use the module we might be able to get a better idea of what the best real-world RUP looks like.

alexis-via:
What do you think of my suggestion of using ssnid instead of identification_id and adding a dependency on l10n_fr_hr_check_ssnid ?

PierrickBrun:
Sorry @alexis-via , I forgot to respond, indeed I think it's a good idea

kevinkhao:
@PierrickBrun it's my bad
@alexis-via sorry for the delay, good idea
I pushed the modification
Thanks

alexis-via (approved):

PR #268, opened 2021-01-05
[14.0][MIG] l10n_fr_intrastat_product

alexis-via:
Requires PR https://github.com/OCA/intrastat-extrastat/pull/116 and https://github.com/OCA/intrastat-extrastat/pull/115 and certainly others.


rvalyi:
@alexis-via it looks the tests fail because of the XML tests in the pre-commit hook: https://github.com/OCA/l10n-france/pull/268/checks?check_run_id=1752025491#step:4:139

May be you should alter pre-commit configuration to skip the validation of l10n_fr_intrastat_product/data/deb.xsd
I'm not too sure, but may be you can try the excludeFiles option in `.prettierrc.yml` see the end of this documentation section https://prettier.io/docs/en/configuration.html#configuration-overrides

PR #269, opened 2021-01-07
[13.0][IMP]capture cfonb transaction details in note field

luc-demeyer:
The current cfonb parser doesn't store sufficient info to allow reconciliation of the bank statement without the printed statement at hand. This PR now stores all transaction details in the bank statement line note field.

### DIFF
--- a/account_bank_statement_import_fr_cfonb/wizard/account_bank_statement_import.py
+++ b/account_bank_statement_import_fr_cfonb/wizard/account_bank_statement_import.py
@@ -171,6 +171,7 @@ class AccountBankStatementImport(models.TransientModel):
                         date_str, ref, amount, name
                     ),
                     "amount": amount,
+                    "note": name,
                 }
                 seq += 1
             elif rec_type == "05":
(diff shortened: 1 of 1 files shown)


alexis-via:
What is the complementary_info_type of your lines that start with 05 ? It is a "proprietary" type ? I guess it's not LIB nor empty, otherwize you would already have it in the name field.

I'm ok to store proprietary complementary_info_type  in the note field, but I think it's a bit strange to also have the content of the name field in the note field. Note field should store additionnal data.

luc-demeyer:
@alexis-via 
I am processing statements of the CIC bank.
The complimentary info that my customer receives via the CIC bank contains basic info such as e.g. payment reference, hence it is an absolute must to include this.
Adding it to the Notes Field together with the 'name' field makes the whole thing a bit more readable since the full transaction detail is stored in the Note field.
I cannot add concrete examples to this PR since the bank statement details of my customer is sensitive data.

alexis-via (approved):

PR #274, opened 2021-02-18
[14.0] l10n_fr_siret: add warning for duplicates

alexis-via:
siren_dup

This warning is similar to the warning on VAT duplicate in the 'base' module


legalsylvain:
Hum, i like the screenshot "Duplicate, Big Burden".

PR #282, opened 2021-03-26
[14.0] DAS2: Avoid error when the amount is negative

alexis-via:


alexis-via:
Up-port of v12 fix https://github.com/OCA/l10n-france/pull/281/files

rlemeur (approved):

PR #289, opened 2021-06-03
[IMP] Add parameter for DEB per invoice

saxomanu:


tcamberlin:
Thanks for this PR.
Il adds a check box in accounting configuration to decide weither we want one DEB line per invoice or to merge all invoices in one DEB line (when applicable, ie same destination country, same HS Code, same commercial partner).
It is very useful for two reasons:
* some people doing this procedure are used to work on a per invoice basis. By having DEB module reproduce this behaviour, we lower the stair when making the shift from X to Odoo.
* it eases controling DEBs : internally or externally by administration.

PR #294, opened 2021-07-13
[FIX] l10n_fr_hr_rup: remove super()._onchange_employee_id()

clementmbr:
The native `_onchange_employee_id()` were converted to stored-editable computed fields in :
https://github.com/odoo/odoo/commit/3bd345597fa932abdaf780f72bacf331690f549b

cc @sebastienbeau @kevinkhao

### DIFF
--- a/l10n_fr_hr_rup/models/hr_contract.py
+++ b/l10n_fr_hr_rup/models/hr_contract.py
@@ -7,14 +7,31 @@ class HrContract(models.Model):
     _inherit = "hr.contract"
     _order = "date_start,name"
 
-    pcs_id = fields.Many2one("hr.employee.pcs", "PCS")
     employer_address_id = fields.Many2one("res.partner", "Addresse de l'employeur")
-    qualification = fields.Char("Qualification")
-    work_location = fields.Char("Localisation du bureau")
+    pcs_id = fields.Many2one(
+        "hr.employee.pcs", "PCS", compute="_compute_pcs_id", store=True, readonly=False
+    )
+    qualification = fields.Char(
+        "Qualification", compute="_compute_qualification", store=True, readonly=False
+    )
+    work_location = fields.Char(
+        "Localisation du bureau",
+        compute="_compute_work_location",
+        store=True,
+        readonly=False,
+    )
 
-    @api.onchange("employee_id")
-    def _onchange_employee_id(self):
-        super()._onchange_employee_id()
-        self.pcs_id = self.employee_id.pcs_id
-        self.qualification = self.employee_id.qualification
-        self.work_location = self.employee_id.work_location
+    @api.depends("employee_id")
+    def _compute_pcs_id(self):
+        for rec in self:
+            rec.pcs_id = rec.employee_id.pcs_id
+
+    @api.depends("employee_id")
+    def _compute_qualification(self):
+        for rec in self:
+            rec.qualification = rec.employee_id.qualification
+
+    @api.depends("employee_id")
+    def _compute_work_location(self):
+        for rec in self:
+            rec.work_location = rec.employee_id.work_location


clementmbr:
Pre-commit failing because of other modules...
https://github.com/OCA/l10n-france/pull/294/checks?check_run_id=3056420286#step:4:103

l10n_fr_hr_rup/models/hr_contract.py
```diff
@@ -14,7 +14,6 @@ class HrContract(models.Model):
 
     @api.onchange("employee_id")
     def _onchange_employee_id(self):
-        super()._onchange_employee_id()
```
ivantodorovich, line 17: **Alternative 1: onchange**

```suggestion
        res = None
        if hasattr(super(), "_onchange_employee_id"):
            res super()._onchange_employee_id()    # pragma: no cover
```
and `return res` at the end.

This is to play nice with other modules that may also be defining an `_onchange_employee_id` method.
The name is generic enough, so one has to be careful.

-----

**Alternative 2: computed stored writeable field** :star2: _Preferred soltution_ 
Convert `pcs_id`, `qualification`, `work_location` into computed fields with `store=True` and `readonly=False`. each with it's own `_compute_$field_name` method depending on `@api.depends("employee_id")`

ivantodorovich (changes requested):

l10n_fr_hr_rup/models/hr_contract.py
clementmbr: Thanks for the comment @ivantodorovich !
Ok to go for alternative 2.
Just one question : why is it better to use 3 different `_compute` methods instead of only one (the same for the 3 fields) ?

ivantodorovich: Technically in this case there isn't any difference. But I find it's better for extendability, so in general it's a good practice.

My rule of thumb is to always use different compute methods, with the only exception being the case were there's a performance gain in grouping.

clementmbr: done ! https://github.com/OCA/l10n-france/pull/294/files#diff-058c69791d50b8616ac786f6e1aa76ae3338117cae45400e9cab2510bd6dc0f8R11

Thanks again for the review !

l10n_fr_hr_rup/models/hr_contract.py
```diff
@@ -7,14 +7,31 @@ class HrContract(models.Model):
...
-    @api.onchange("employee_id")
-    def _onchange_employee_id(self):
-        super()._onchange_employee_id()
-        self.pcs_id = self.employee_id.pcs_id
-        self.qualification = self.employee_id.qualification
-        self.work_location = self.employee_id.work_location
+    @api.depends("employee_id")
+    def _compute_pcs_id(self):
+        for contract_id in self:
+            contract_id.pcs_id = self.employee_id.pcs_id
```
ivantodorovich:
```suggestion
        for contract_id in self:
            contract_id.pcs_id = contract_id.employee_id.pcs_id
```

Also change the others please.
Otherwise you get an error when recomputing multiple records




Also, as a non blocking suggestion, I prefer naming the single record `rec` or `record`, as it's pretty common to find it like this and it's a "standard" that applies to any model

```suggestion
        for rec in self:
            rec.pcs_id = rec.employee_id.pcs_id
```

ivantodorovich (changes requested):

l10n_fr_hr_rup/models/hr_contract.py
clementmbr: Ops, sorry for this mistake and thanks again for the advises!

ivantodorovich (approved):

PierrickBrun (approved):

l10n_fr_hr_rup/models/hr_contract.py
```diff
@@ -7,14 +7,31 @@ class HrContract(models.Model):
...
+        for rec in self:
+            rec.pcs_id = rec.employee_id.pcs_id
+
+    @api.depends("employee_id")
+    def _compute_qualification(self):
+        for rec in self:
+            rec.qualification = rec.employee_id.qualification
+
+    @api.depends("employee_id")
+    def _compute_work_location(self):
```
kevinkhao, line 35: I think this kind of information is susceptible to change; by computing it we lose historic data on the contracts

l10n_fr_hr_rup/models/hr_contract.py
```diff
@@ -7,14 +7,31 @@ class HrContract(models.Model):
...
-        self.pcs_id = self.employee_id.pcs_id
-        self.qualification = self.employee_id.qualification
-        self.work_location = self.employee_id.work_location
+    @api.depends("employee_id")
+    def _compute_pcs_id(self):
+        for rec in self:
+            rec.pcs_id = rec.employee_id.pcs_id
+
+    @api.depends("employee_id")
+    def _compute_qualification(self):
```
kevinkhao, line 30: Same here, an employee might change qualification over time

kevinkhao (changes requested):

l10n_fr_hr_rup/models/hr_contract.py
kevinkhao: Sorry, that's not actually the case since it is computed with store=True. This makes it more user-friendly but also easier to make mistakes IMHO

ivantodorovich: These are computed stored writeable fields. From the user perspective / functional side, there's no difference with the onchange.

Can you elaborate on your concerns?

kevinkhao: Yes, that's my mistake. I thought about some client making a mistake by changing employee_id but thinking about it that's out of the scope of this PR, I switch to approve now to avoid blocking

kevinkhao (approved):

PR #298, opened 2021-08-10
[REF][14.0] Fix CI pre-commit + travis.yml, ignoring modules

legalsylvain:
ignoring the following modules thats makes travis / runbot red on 12.0 branch 

- l10n_fr_account_invoice_import_facturx
- l10n_fr_business_document_import
- l10n_fr_intrastat_product
- l10n_fr_intrastat_service
- account_balance_ebp_csv_export
- account_banking_fr_lcr
- account_statement_import_fr_cfonb
- l10n_fr_account_tax_unece
- l10n_fr_fec_oca
- l10n_fr_mis_reports

PR #302, opened 2021-09-01
[14.0] Restore tests and remove Alpha status

alexis-via:
This is the "real" work to make CI work on OCA/l10n-france. It aims at cancelling the "Donald Trump PR" #298 . Most of the changes are trivial. On l10n_fr_fec_oca, it's unfortunately not easy to conform it to pylint; it would require diverging even more from l10n_fr_fec which would make comparisons more difficult and code exchange between the 2 modules would become a big pain. That's why I disabled pylint for the moment on the file wizard/account_fr_fec_oca.py

This PR also ends the humiliation of module authors/contributors who saw some 10-year-old module, born at the time of Odoo v5, used in production by a lot of companies, suddenly flagged as Alpha software with the mention in the README "This is an alpha version, the data model and design can change at any time without warning. Only for development or testing purpose, do not use in production." Humiliating module authors and contributors is not something we should tolerate in OCA/l10n-france. As PSC, I'll fight against this kind of attitude in OCA/l10n-france.


legalsylvain (commented):
Hi @alexis-via.

If you want to make travis + runbot install and check modules you should also remove module names from the list here : 
https://github.com/OCA/l10n-france/blob/14.0/.travis.yml#L30

the ``.pre-commit-config.yaml`` is just about syntax. (pylint, flake8, ...)


BTW : nice commit message ;-)

alexis-via:
Done. Green travis.

ivantodorovich (approved):
LG :+1:

legalsylvain (approved):
Thanks !

PR #303, opened 2021-09-01
[12.0] Restore tests and remove Alpha status

alexis-via:
This is the "real" work to make CI work on OCA/l10n-france. It aims at cancelling PR #297 . 

This PR also ends the humiliation of module authors/contributors who suddenly saw their modules flagged as Alpha software without notice with the mention in the README "This is an alpha version, the data model and design can change at any time without warning. Only for development or testing purpose, do not use in production." Humiliating module authors and contributors is not something we should tolerate in OCA/l10n-france. As PSC, I'll fight against this kind of attitude in OCA/l10n-france.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,11 @@ addons:
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
-      - python-lxml  # because pip installation is slow
-      - python-simplejson
-      - python-serial
-      - python-yaml
+      - python3-oauthlib  # because pip installation is slow
 
 env:
   global:
   - VERSION="12.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0"
-  - EXCLUDE=l10n_fr_chorus_facturx,l10n_fr_chorus_ubl,l10n_fr_business_document_import,l10n_fr_chorus_account,l10n_fr_fec_oca,l10n_fr_chorus_sale
   matrix:
   - LINT_CHECK="1"
   - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
--- a/l10n_fr_business_document_import/__manifest__.py
+++ b/l10n_fr_business_document_import/__manifest__.py
@@ -8,7 +8,7 @@
     'license': 'AGPL-3',
     'summary': 'Adapt the module base_business_document_import for France',
     'author': 'Akretion,Odoo Community Association (OCA)',
-    'development_status': 'Alpha',
+    'maintainers': ['alexis-via'],
     'website': 'https://github.com/OCA/l10n-france',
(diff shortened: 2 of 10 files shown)


legalsylvain (approved):
Thanks. All is now green, except low coverage (36%), but not a blocking point.

:+1:

ivantodorovich (approved):

rvalyi (approved):

alexis-via:
Disabling tests and degrading modules to Alpha status: 7 lines of code (cf PR https://github.com/OCA/l10n-france/pull/297/files)
Doing the real work: 4 lines of code

pedrobaeza:
Well, @alexis-via please let's move on. This is like the joke of unfolding the concepts of an inflated bill about tightening a screw:

- Tighten a screw - 5 €
- Know which screw to tighten - 1000 €

For you as the author is simpler to know where to act "surgically", but the general approach was to be safe-guard.

Anyways, there's not too much difference between Alpha and Beta IMO. Promote the modules to "Development/Stable" or "Mature" if you consider so.

alexis-via:
What I really dislike is the spirit behind this: instead of helping fix the problem, some contributors prefer humiliating module authors by degrading their modules to alpha status without their consent in less than 24h in the middle of August. And it works: in the end, the humiliated guy wants to restore it's module status and does the real work. But it's not my vision of how OCA should work. On the 4 lines of code required to fix the real problem, 3 of them were trivial and didn't require any prior knowledge of the codebase; only 1 of them required some knowledge of the codebase.

Don't worry, I'll move on. But the lesson I learn from this is that, on OCA/l10n-france, I can't count on any help to do the real work. And being the main contributor, the main author of modules and PSC of OCA/l10n-france doesn't protect you from public humiliation on your hard work. I don't expect to receive "thank you" for my work here, but I didn't imagine I could receive this kind of thing.

pedrobaeza:
Alexis, I don't think there's any spirit about humiliating you. We all 3 know each other and that's not the style of anyone. A bit upset for not following guidelines, yeah, and you know I have been in the same situation with you in other contributions, but nothing to degrade to that level. It was more the uncertainty to know how the modules work, and he didn't expect nor want to provoke such effect.

About the loneliness on localization maintenance, I have been in such situation for a long time (now it has got better) in OCA/l10n-spain, but you can't be part of the problem making harder to start contributing. Take the reverse: if this move has been done, it's because there's interest in starting to contribute/maintain.

I reserve an appointment for the next physical Odoo XP or OCA days (hopefully on 2022!) to take a beer the 3 of us remembering this thing as an anecdote.

alexis-via:
ok :) Like you, I miss our regular in-person OCA code sprints and also Odoo XP.

PR #304, opened 2021-09-02
[14.0][FIX] l10n_fr_hr_rup: fix demo data with invalid ssnid

kevinkhao:
demo data didn't comply with ssnid verification


hparfr (changes requested):
Since you are at it, please add valid numbers.
If it's a male, start with 1 then compute the key % 97.

hparfr (approved):
LGTM
Thanks :nail_care:

PR #309, opened 2021-10-08
[14.0][FIX] account_balance_ebp_csv_export: missing quote before account code

alexis-via:
Found the bug while testing on teledec.fr.


legalsylvain:
Make sense. fast-tracking trivial PR.

/ocabot merge patch

PR #173, opened 2019-02-18
[MIG][12.0] l10n_fr_siren_lookup

remi-filament:
Port Siren Lookup module from v10.0 to v12.0, updating also SIRENE API from v1 to v3

PR #310, opened 2021-10-14
[14.0] l10n_fr_siret: set siret field RW and use stdnum for checksum control

alexis-via:
The auto-complete feature is made by an onchange on name, siren and siret

Field 'siret' is now readonly=False on res.partner, to make it easier to copy/paste a siret from an external source

Add tests

The PR implement a part of what has been discussed with Remi from Le Filament during today's meeting announced on https://github.com/OCA/l10n-france/pull/173

Left to do:
- add VAT number in autocomplete via compute of checksum + check on VIES
- add methods to create a new partner from siren/siret

@remi-filament


l10n_fr_siret/models/res_partner.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+def _check_luhn(string):
+    """Luhn test to check control keys
+
+    Credits:
+        http://rosettacode.org/wiki/Luhn_test_of_credit_card_numbers#Python
+    """
+    r = [int(ch) for ch in string][::-1]
+    return (sum(r[0::2]) + sum(sum(divmod(d * 2, 10)) for d in r[1::2])) % 10 == 0
```
legalsylvain: the luhn check seems available in the following lib : 
https://arthurdejong.org/python-stdnum/doc/1.4/stdnum.luhn.html

(luhn check seems used by odoo, in V14).

l10n_fr_siret/models/res_partner.py
```diff
@@ -0,0 +1,388 @@
+import logging
```
legalsylvain, line 1: the git rename of this file failed. (for company it's ok).
so the diff review is not easy to do.

legalsylvain (commented):
hi @alexis-via. First thanks for this very great feature ! 

Some minor comment inline and two more important points.

**Regarding the module design**

- you are introducing a new feature that are making call to external API ;
- you add an external python dependencies to requests. (quite common but well)
- there is no option to disable the api call. (if I understood correctly, but maybe I missed something)
- if the call fails, or if data are not found, code generate warning logs.

so, I'm not sure that making such changes in an existing and deployed module is ok. moreover it is marked as "mature". (l10n_fr_siret module is a nice to have module for all french companies.)
An alternative could be to create another module named ``l10n_fr_siret_opendatasoft_autocomplete`` to implement that feature. What do you think ?

**Regarding the feature design** 
I'm not very confortable with the onchange. I'm not sure on that point, but I see it quite intrusive. If I understand correctly, when user set siret, it will change automatically name, street, etc.. dependending on opendatasoft data. Well, maybe in 80%, it will help, but in 20%, previous data was correct and new ones are not. (wrong data, data not up to date, etc...) So, I'm not sure what could be the best design, but something like a button, calling a wizard, displaying old and new data, could be more explicit for end user.

Don't you think ?
On that second point, if you're agree with the popup wizard proposal, but if you have no time, I can try to make a PoC.

kind regards.

l10n_fr_siret/models/res_partner.py
```diff
@@ -0,0 +1,388 @@
...
+        size=9,
+        tracking=50,
+        help="The SIREN number is the official identity "
+        "number of the company in France. It composes "
+        "the first 9 digits of the SIRET number.",
+    )
+    nic = fields.Char(
+        string="NIC",
+        size=5,
+        tracking=51,
```
legalsylvain: Just a question but what is this syntax ? I dont know it. I usually use tracking=True.

remi-filament:
Wow, thanks Alexis for the fast implementation !
Thanks Sylvain for review, I understand your points but may I propose some more insights:
From what we discussed with Alexis and has been (super fast) implemented in code :
- allow user to enter Siret directly instead of having to split siren and nic to calculate siret (most of my users have Siret directly so it's easier not to ask them to split it) : this was added with inverse fonction and making Siret field readwrite
- in case user enters siren or Siret directly and only if no other field is filled (no name, no address) : the request is made to fill thèse
- you can enter Siret or Siren directly in name field, and then only in that case (9 or 14 digits in name field) and other fields are empty (siren, nic, Siret, address) thé request is made to fill the fields

I guess that with the above explanations you would agree that unless user applies these very specific use cases (which I think he cannot guess unless well informed), no request would be sent to opendatasoft and in any case request and filling fields would only be performed if all other fields are empty.

Let us know if with these extra explanations you still think it should be done another way or elsewhere ?

Thanks !

l10n_fr_siret/models/res_partner.py
alexis-via: It configures the order of the field in the change message in the chatter.

legalsylvain: nice ! thanks for the info !

legalsylvain:
Hi. point to point : 

> allow user to enter Siret directly instead of having to split siren and nic to calculate siret (most of my users have Siret directly so it's easier not to ask them to split it) : this was added with inverse fonction and making Siret field readwrite

yes, this is a specific point of this PR. I miss this feature also in V12, and was thinking to implement it also. I'm 100% ok with that.

> you can enter Siret or Siren directly in name field, and then only in that case (9 or 14 digits in name field) and other fields are empty (siren, nic, Siret, address) thé request is made to fill the fields

That use case seems ok for me indeed. Thanks for the clarification I didn't understood that. it is the classical "Create" use case. no previous data erased : :+1: 

> in case user enters siren or Siret directly and only if no other field is filled (no name, no address) : the request is made to fill thèse

This is the classical "Update" use case. (partner still exists). In that case, I still think that having the following implementation could be more user friendly : 
- a button near the siret field "Update from Opendatasoft"
- a wizard displayed when clicking on that button displaying old / new info

-> It really shows to the user the diff between current data and futur data.
-> It allows to execute the update of the data, even if the fields are not empty. (name, adress, etc...) that is in my opinion, usefull for existing database with a lot of customers / suppliers

Last point, regarding the stability of the module, what do you think ? I mean this module is available on V14 instance since [2020 December](https://github.com/OCA/l10n-france/pull/252#event-4086138083). and this PR introduces new dependencies. Making a new module solve that problem. and allow user to select if they want to install or not such extra features.
My PoV is that it doesn't hurt to split into two modules, and make it more generic. ``l10n_fr_siret`` could stay "Mature" and ``l10n_fr_autocomplete_opendatasoft`` could be in "Beta" state, as it is quite new, and APi can change, etc...

remi-filament:
Thanks Sylvain,

regarding your point for updating data : 

> This is the classical "Update" use case. (partner still exists). In that case, I still think that having the following implementation could be more user friendly :
> - a button near the siret field "Update from Opendatasoft"
> - a wizard displayed when clicking on that button displaying old / new info

> -> It really shows to the user the diff between current data and futur data.
> -> It allows to execute the update of the data, even if the fields are not empty. (name, adress, etc...) that is in my opinion, usefull for existing database with a lot of customers / suppliers

For me what is proposed here is not an update use case, since for creating a contact you need a name, so we are still in "Create" use case, where instead of looking for SIREN or SIRET fields hidden below, you could directly enter these in name field (and only this information on contact creation) to have other fields (name, siren, nic, siret, address) autocompleted.

We discussed yesterday about implementing some more in another module :
- to search by name or by other fields than SIREN / SIRET
- to propose to update data if already present ("Update" use case)
- to fill in other fields (such as APE code and name) if fields are present from other modules

Regarding the way it works we were thinking about using a wizard (like done in #173 ) or try to make something similar to partner_autocomplete from Odoo for listing possible companies when entering info in name field (although I am personnally much less comfortable with JS...)

Anyway, I have no opinion whether the first part proposed by Alexis should be in l10n_fr_siret or not (since I am not using it today for my customers), so I let you and whoever is interested decide on this.

remi-filament:
Thanks Sylvain,

regarding your point for updating data : 

> This is the classical "Update" use case. (partner still exists). In that case, I still think that having the following implementation could be more user friendly :
> - a button near the siret field "Update from Opendatasoft"
> - a wizard displayed when clicking on that button displaying old / new info

> -> It really shows to the user the diff between current data and futur data.
> -> It allows to execute the update of the data, even if the fields are not empty. (name, adress, etc...) that is in my opinion, usefull for existing database with a lot of customers / suppliers

For me what is proposed here is not an update use case, since for creating a contact you need a name, so we are still in "Create" use case, where instead of looking for SIREN or SIRET fields hidden below, you could directly enter these in name field (and only this information on contact creation) to have other fields (name, siren, nic, siret, address) autocompleted.

We discussed yesterday about implementing some more in another module :
- to search by name or by other fields than SIREN / SIRET
- to propose to update data if already present ("Update" use case)
- to fill in other fields (such as APE code and name) if fields are present from other modules

Regarding the way it works we were thinking about using a wizard (like done in #173 ) or try to make something similar to partner_autocomplete from Odoo for listing possible companies when entering info in name field (although I am personnally much less comfortable with JS...)

Anyway, I have no opinion whether the first part proposed by Alexis should be in l10n_fr_siret or not (since I am not using it today for my customers), so I let you and whoever is interested decide on this.

alexis-via:
I implemented "add VAT number in autocomplete via compute of checksum + check on VIES" (I haven't pushed the code yet). It works well, but it adds latency because we now make 2 requests: 1 on opendatasoft and 1 on VIES. So I thought: "why not get all the data from VIES" ? I knew that this kind of feature was already available in OCA ; I eventually found the module under its new name "partner_data_vies_populator" ; the old name was partner_create_by_vat. To test/analyse the module, I migrated it to v14 cf https://github.com/OCA/partner-contact/pull/1152 In fact, the main pb with this module is that the address is not structured ; there is only 1 line that gives the street + city + zip. In fact, we get 3 infos from VIES : company name, country code and one-line address. So I'm back to my original idea to combine the 2 APIs opendatasoft to have a structured address and VIES for the VAT number. Another advantage is that it will also work for auto-entrepreneurs or others entities without VAT number.

I agree that, if we implement this in l10n_fr_siret we need a way to disable the request to opendatasoft. So let's implement the API part in l10n_fr_siret_lookup. I'll move the code there.

legalsylvain:
Nice ! Could you check travis ? L10n_product_intrastat has bad demo data with bad siret, if i understand correctly the log. 
Otherwise. 👍

alexis-via:
Travis is now green

l10n_fr_siret/models/res_partner.py
```diff
@@ -31,14 +28,32 @@ def _compute_siret(self):
...
+    def _inverse_siret(self):
+        for rec in self:
+            siret = rec.siret and rec.siret.replace(" ", "") or False
+            if siret and len(siret) == 14 and siret.isdigit():
+                rec.write(
+                    {
+                        "siren": siret[:9],
+                        "nic": siret[9:],
+                    }
+                )
```
remi-filament, line 40: You are using here isdigit() when you use isdecimal() everywhere else, I guess we should harmonize ?
Also, should not it be useful to check if siret has incorrect length or format ? (I had to make a special tweak to handle the case where generated from SIREN only with ***** at the end)

```suggestion
            if siret and len(siret) == 14:
                if siret.isdecimal():
                    rec.write(
                        {
                            "siren": siret[:9],
                            "nic": siret[9:],
                        }
                    )
                elif not siret.endswith("*****"):
                    raise ValidationError(
                        _(
                            "The SIRET '%s' is incorrect: it must have "
                            "exactly 14 digits."
                        )
                        % siret
                    )
            elif siret and len(siret) != 14:
                raise ValidationError(
                    _(
                        "The SIRET '%s' is incorrect: it must have "
                        "exactly 14 digits."
                    )
                    % siret
                )
```

remi-filament: Thanks for this update, except for the above detail, it looks much better for my use cases !

legalsylvain (approved):
Thanks for this great addition.

l10n_fr_siret/models/res_partner.py
alexis-via: I harmonized isdigit vs isdecimal
On the second point, no change needed in the code because the _check_siret() method will not be by-passed even if you go through _inverse_siret(). The inverse method is just another way to set siren/nic ; the @api.constrain on siren/nic will continue to apply.

remi-filament: Actually if the Siret you enter ils not 14 char long or only digit it will not raise any error/warning for invalid Siret since it will not set siren/nic and thus will not check Siret neither.This is why I proposed to add some checks to warn user about incorrect Siret.
For instance if you forget one 0 in the NIC part you will not get warned.

Since this module is checking length / format for siren and NIC I would have expected that it would do the dame for siret

PR #313, opened 2021-10-25
[14.0][FIX] l10n_fr_siret: fix bad data when entering wrong SIRET

alexis-via:
@remi-filament I fixed it keeping the code as simple as possible.


remi-filament (approved):
Great thanks !

PR #315, opened 2021-10-26
[14.0][FIX] partner nic validation when changing siren

Kev-Roche:
When a contact is switching from a partner with siren and nic numbers, to another partner with different siren and nic numbers, validation of the siret failed (function  _check_siret). 

The Validation error is about the nic which is update after the siren. So the _check_siret() takes into account the new siren and the original nic, bringing an invalid checksum. 

I added the corresponding test and a working proposition to fix it.

### DIFF
--- a/l10n_fr_siret/models/res_partner.py
+++ b/l10n_fr_siret/models/res_partner.py
@@ -44,6 +44,8 @@ class Partner(models.Model):
     def _check_siret(self):
         """Check the SIREN's and NIC's keys (last digits)"""
         for rec in self:
+            if rec.type == "contact" and rec.parent_id:
+                continue
             if rec.nic:
                 # Check the NIC type and length
                 if not rec.nic.isdigit() or len(rec.nic) != 5:
--- a/l10n_fr_siret/tests/test_fr_siret.py
+++ b/l10n_fr_siret/tests/test_fr_siret.py
@@ -108,3 +108,32 @@ class TestL10nFrSiret(TransactionCase):
         self.assertFalse(partner_company2.same_siren_partner_id)
         partner_company2.write({"company_id": False})
         self.assertEqual(partner_company2.same_siren_partner_id, partner_company1)
+
+    def test_change_parent_id(self):
+        partner = self.env["res.partner"].create(
+            {
+                "name": "Akretion France",
+                "siren": "792377731",
+                "nic": "00023",
+            }
+        )
+
+        partner2 = self.env["res.partner"].create(
+            {
+                "name": "Akretion Fr",
+                "siren": "784671695",
+                "nic": "00087",
+            }
+        )
+
+        contact = self.env["res.partner"].create(
+            {"name": "Test contact", "parent_id": partner.id}
+        )
+
+        self.assertEqual(partner.siren, contact.siren)
+        self.assertEqual(partner.nic, contact.nic)
+
+        contact.write({"parent_id": partner2.id})
+
+        self.assertEqual(partner2.siren, contact.siren)
+        self.assertEqual(partner2.nic, contact.nic)


Kev-Roche:
cc @sebastienbeau, @alexis-via what do you think about it, maybe you will think about a more elegant way to fix it?

legalsylvain:
> When a contact is switching from a partner with siren and nic numbers, to another partner with different siren and nic numbers, validation of the siret failed (function _check_siret).

I'm not sure it's valid. 
i mean 00001 is a very common nic.

Use Case : I have a customer / supplier with SIREN A + NIC : 00001
My customer / supplier say me the siret changed.
So I update the siren field with SIREN B.
With your code, the nic will be erased.
is it desired ?

Kev-Roche:
> > When a contact is switching from a partner with siren and nic numbers, to another partner with different siren and nic numbers, validation of the siret failed (function _check_siret).
> 
> I'm not sure it's valid. i mean 00001 is a very common nic.
> 
> Use Case : I have a customer / supplier with SIREN A + NIC : 00001 My customer / supplier say me the siret changed. So I update the siren field with SIREN B. With your code, the nic will be erased. is it desired ?

@legalsylvain Yes it is desired. 
From my perspective, if the siren changes, the nic also needs to change to keep the checksum of the siret correct (Luhn algorithm).

legalsylvain:
OK. I get it. I didn't know that in NIC, there was also a checksum digit.
well, I still don't think overwriting write is a good idea. I mean, a constrains is enought to check the validity. if the user doesn't update the nic, he receive a usererror that is ok.

don't you think ?

Kev-Roche:
I'm not sure that overwriting write method is good here, but I did not find better idea to fix it. 
The problem is that the validation error fires automatically when a contact changes its parent_id which already have a siren + nic. In that case, propagation of the siren is working well from parent to contact, then checks occur, then after that nic is propagate to the adress. So when the siret check occurs, the siren and nic are not from the same parent_id. Consequently validation error occurs for a wrong reason in my opinion.

alexis-via:
Changing a contact from a parent partner to another is extremely dangerous, I would recommend to archive the old contact and create a new one. One of the reasons: if you had a document (order, invoice) linked to that contact and that the commercial_partner_id field on this document is a related field, then the document will be linked to the new company! This is certainly not what you want. I had a look at the code for the invoice (https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_move.py#L1323), it seems it's done right in v14 because it's not a related field but a computed field that a proper invalidation, but it's not done that way in sale_commercial_partner nor in purchase_commercial_partner (we should probably update the code of those modules by the way). And it was not done the right way on invoices in v12 for example.

l10n_fr_siret/models/res_partner.py
```diff
@@ -50,6 +50,8 @@ def _inverse_siret(self):
     def _check_siret(self):
         """Check the SIREN's and NIC's keys (last digits)"""
         for rec in self:
+            if rec.type == "contact" and rec.parent_id.id != False:
```
sebastienbeau:
```suggestion
            if rec.type == "contact" and rec.parent_id:
```

sebastienbeau:
I agree that it's not the best idea to change a contact of company.
But blocking this behavior should be done in a dedicated module and not by a random error (you can change the company of a contact if the company do not have a siret)

Also in our case the aim is not to change the contact because the contact have a new work in a new company.
Our aim is to solve an incompatibility with the module "intercompany_shared_contact"
PR here : https://github.com/OCA/multi-company/pull/315
Indeed this module will 
- fill the parent_id of res_partner linked to res.users (we fill it with the company partner)
- adapt access rule so contact are visible by all company.

The issue is with the parent_id, if you try to change the company_id of a user you can't because you have the nic check error.

As checking the nic on a contact that have a parent_id is useless, I think we can merge this and see in an other PR if we want to block the modification of the parent_id in some case (but this should be done in a different module).

sebastienbeau (approved):

sebastienbeau:
@alexis-via @legalsylvain

legalsylvain (approved):

Kev-Roche:
@legalsylvain , @sebastienbeau, thanks for the approval, I just fixed the pre-commit problem.

PR #320, opened 2021-11-19
[13.0] l10n_fr_intrastat_service: add migration script

alexis-via:
required for v12 to v13 mig


legalsylvain (approved):
Make sense. 
Thanks. 
Could you check travis ?

alexis-via:
The origin/13.0 branch of l10n-france is not green. The pb is not in this PR. We would need to make it green as I did for 12.0 and 14.0.

PR #326, opened 2021-12-12
[14.0] DEB: adapt for January 1st 2022 reform

alexis-via:
I propose that we DON'T merge this before January 10th 2022 (the reform starts on DEB of January 2022 that must be done in early February 2022).


alexis-via:
Same PR for odoo v10 https://github.com/OCA/l10n-france/pull/327
Same PR for odoo v12 https://github.com/OCA/l10n-france/pull/319

PR #329, opened 2021-12-20
[15.0][MIG] l10n_fr_ecotaxe

SilvioC2C:

PR #330, opened 2021-12-21
[15.0][IMP] l10n_fr_ecotaxe: manage ecotax on products

SilvioC2C:
Allows managing ecotax functionalities directly on products instead of templates.
This way, a variant can have an ecotax configuration that is different from its template's configuration.

Includes and depends on https://github.com/OCA/l10n-france/pull/329


ivantodorovich:
Maybe this functionality could be split in a separate module instead?

Because default odoo behaviour for taxes is to set them on the template, not on variants. I think it's a good idea to keep ecotaxes aligned to that, no?

SilvioC2C:
Moved to #336

SilvioC2C:
Moved to #336

PR #331, opened 2021-12-31
[MIG][15.0] l10n_fr_siret

alexis-via:


l10n_fr_siret/models/res_partner.py
```diff
@@ -0,0 +1,167 @@
...
+        inverse="_inverse_siret",
+        store=True,
+        readonly=False,
+        help="The SIRET number is the official identity number of this "
+        "company's office in France. It is composed of the 9 digits "
+        "of the SIREN number and the 5 digits of the NIC number, ie. "
+        "14 digits.",
+    )
+    company_registry = fields.Char(
+        size=64,
```
legalsylvain: Does it makes sense ? I think we could remove that limitation.

legalsylvain (approved):
Diff review. A minor comment.
Lgtm otherwise. Thanks !
Do you know why there a lot of empty translation files ?
Maybe we could remoce them ? 
Regards and happy new year.

l10n_fr_siret/models/res_partner.py
alexis-via: ok, I removed it

PR #335, opened 2022-01-02
[15.0][MIG] account_banking_fr_lcr

alexis-via:


account_banking_fr_lcr/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+{
+    "name": "French Letter of Change",
+    "summary": "Create French LCR CFONB files",
+    "version": "15.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Akretion,Odoo Community Association (OCA)",
+    "maintainers": ["alexis-via"],
+    "website": "https://github.com/OCA/l10n-france",
+    "category": "French localisation",
+    "depends": ["account_payment_order"],
```
ajaniszewska-dev, line 15: travis fails because it cannot find account_payment_order in this version, if it is not migrated yet, but three is PR with migration, add dependency to this PR inside test-requirements.txt file.

PR #339, opened 2022-01-28
[15.0][MIG] DES

alexis-via:


legalsylvain (approved):
Diff review.
LGTM.
Thanks.

legalsylvain:
Hi. For your information CI is red because the module ``instrastat_base`` is not available. (https://github.com/OCA/intrastat-extrastat/pull/157 is not merged)
OCA is removing Travis and now uses github actions.
- OCA merged modules are now automatically found without any actions, via pypi (that's great!)
- OCA unmerged modules like ``instrastat_base`` can be referenced temporary in the test-requirements.txt file in the root folder.

Detail of the syntax available here : https://github.com/OCA/oca-ci/issues/23#issuecomment-1016645879

regards.

alexis-via:
If someone knows how we can force Travis to install a decent version of stdnum, it would allow to merge this PR, which would be great...

alexis-via:
rebased

alexis-via:
Seems to be caused by the fact that requirements.txt from odoo v15 asks for python-stdnum 1.13 and 1.13 doesn't have vatin :-( I see no other reason.

PR #271, opened 2021-02-04
14.0 mig l10n fr ecotaxe

mourad-ehm:

PR #344, opened 2022-02-09
15.0 mig l10n fr ecotaxe sale

ajaniszewska-dev:
reuses: https://github.com/OCA/l10n-france/pull/271


test-requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/OCA/l10n-france@refs/pull/329/head#subdirectory=setup/l10n_fr_ecotaxe
```
ivantodorovich: ⚠️ DO NOT MERGE this PR without removing this commit first

I suggest to add a message in the PR description as well, and linking the depended PR too

l10n_fr_ecotaxe_sale/tests/test_sale_ecotaxe.py
```diff
@@ -0,0 +1,92 @@
...
+from odoo.addons.account.tests.common import AccountTestInvoicingCommon
+
+
+class TestsaleEcotaxe(AccountTestInvoicingCommon):
+    @classmethod
+    def setUpClass(cls):
+        super(TestsaleEcotaxe, cls).setUpClass()
+
+        cls.ecotaxe_classification = cls.env["account.ecotaxe.classification"]
+        cls.ecotaxe_classification1 = cls.ecotaxe_classification.create(
```
ivantodorovich, line 17: It's weird, but looking at codecov comments it seems tests aren't running for this module

```
INFO odoo unittest.suite: skipped setUpClass (odoo.addons.l10n_fr_ecotaxe_sale.tests.test_sale_ecotaxe.TestsaleEcotaxe)
```

Looking into this I think this is coming from `AccountTestInvoicingCommon`:

https://github.com/odoo/odoo/blob/c8f7a427b9b9fcdc857a71f99f8448f4d19c9a60/addons/account/tests/common.py#L38

I believe you have to pass the chart of accounts xmlref when calling super, to get this working.

ivantodorovich (changes requested):

l10n_fr_ecotaxe_sale/tests/test_sale_ecotaxe.py
ajaniszewska-dev: @ivantodorovich updated code with argument to fix it.

ivantodorovich: Unfortunately tests are still not running.

You can see it in the CI logs:

```
skipped setUpClass (odoo.addons.l10n_fr_ecotaxe_sale.tests.test_sale_ecotaxe.TestsaleEcotaxe) 
```


I think you need to do something like this, based on existing odoo code:

```python
    @classmethod
    def setUpClass(cls, chart_template_ref="l10n_fr.l10n_fr_pcg_chart_template"):
        super(TestsaleEcotaxe, cls).setUpClass(chart_template_ref)
```

ref : https://github.com/odoo/odoo/blob/0eaa4c0ece9927c80c5e2e1991d49093a8f774cc/addons/l10n_fr_fec/tests/test_wizard.py#L18

l10n_fr_ecotaxe_sale/views/sale_view.xml
```diff
@@ -0,0 +1,30 @@
...
+   Copyright (C) 2015-2020 Akretion (http://www.akretion.com/)
+   @author: Mourad EL HADJ MIMOUNE <mourad.elhadj.mimoune@akretion.com>
+-->
+<odoo>
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">l10n_fr_ecotaxe.sale.line.form</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <!-- <field name="amount_untaxed"> -->
```
ivantodorovich:
```suggestion
```
Better to remove this comment if it's not used

l10n_fr_ecotaxe_sale/tests/test_sale_ecotaxe.py
ivantodorovich: Good news: Tests are running now !
Bad news: They're broken 😅

ajaniszewska-dev: hmm interesting. it complains about create_sale_partner and i do not see it defined anywhere in previous migrations too... I am on it :)

l10n_fr_ecotaxe_sale/views/sale_view.xml
ajaniszewska-dev: I kept it to find better place to put this field on. Previously it was under amount_untaxed but there is no such field anymore. I guess it should go into total price section under so lines. Comment will be removed as soon as I'll move this field into better place.

l10n_fr_ecotaxe_sale/tests/test_sale_ecotaxe.py
ajaniszewska-dev: @ivantodorovich  i cannot really run tests locally because they fails on this reference, without that they are skipped:
ValueError: External ID not found in the system: l10n_fr.l10n_fr_pcg_chart_template

i've added dependency to l10n_fr, but no effect. Any idea?

ivantodorovich: I'm not sure if it helps, but maybe try with this: `@tagged('post_install_l10n', 'post_install', '-at_install')`


https://github.com/odoo/odoo/blob/efcb432f4c459b2d2d819c3aed7d63d53fc93809/addons/l10n_fr_fec/tests/test_wizard.py#L14

ajaniszewska-dev: tried it already but no effect. Will keep fixing & pushing here, it doesn't take much time to run the tests and i am not so far from finishing.

l10n_fr_ecotaxe_sale/tests/test_sale_ecotaxe.py
```diff
@@ -0,0 +1,124 @@
...
+        self.sale_line2 = self.sale.order_line[1]
+        self.sale_line1._compute_ecotaxe()
+        self.assertEqual(self.product_a.ecotaxe_amount, 2.4)
+        self.sale_line1.product_uom_qty = 3
+        self.sale_line1.product_uom_change()
+        self.assertEqual(self.sale_line1.unit_ecotaxe_amount, 2.4)
+        self.assertAlmostEqual(self.sale_line1.subtotal_ecotaxe, 7.2)
+        self.assertEqual(self.sale_line2.unit_ecotaxe_amount, 16)
+        self.assertEqual(self.sale_line2.subtotal_ecotaxe, 32)
+        self.assertEqual(self.sale.amount_total, 4300.0)
```
ajaniszewska-dev, line 123: @ivantodorovich  i am bit worried about assertions here, in this tc we have 2 sale order lines for product_a and product_b with qty 1 and 3, on each line product price is 100.00, don't get the point where total 3k - 4k comes from? Is it because of the weight computation for ecotax?

l10n_fr_ecotaxe_sale/tests/test_sale_ecotaxe.py
```diff
@@ -0,0 +1,124 @@
...
+        cls.ecotaxe_classification2 = cls.ecotaxe_classification.create(
+            {
+                "name": "Weight based",
+                "ecotaxe_type": "weight_based",
+                "ecotaxe_coef": 0.04,
+                "ecotaxe_product_status": "P",
+                "ecotaxe_supplier_status": "FAB",
+            }
+        )
+        cls.product_a.weight = 100
```
ajaniszewska-dev, line 35: product_a and product_b comes from account test:
```

        cls.product_a = cls.env['product.product'].create({
            'name': 'product_a',
            'uom_id': cls.env.ref('uom.product_uom_unit').id,
            'lst_price': 1000.0,
            'standard_price': 800.0,
            'property_account_income_id': cls.company_data['default_account_revenue'].id,
            'property_account_expense_id': cls.company_data['default_account_expense'].id,
            'taxes_id': [(6, 0, cls.tax_sale_a.ids)],
            'supplier_taxes_id': [(6, 0, cls.tax_purchase_a.ids)],
        })
        cls.product_b = cls.env['product.product'].create({
            'name': 'product_b',
            'uom_id': cls.env.ref('uom.product_uom_dozen').id,
            'lst_price': 200.0,
            'standard_price': 160.0,
            'property_account_income_id': cls.copy_account(cls.company_data['default_account_revenue']).id,
            'property_account_expense_id': cls.copy_account(cls.company_data['default_account_expense']).id,
            'taxes_id': [(6, 0, (cls.tax_sale_a + cls.tax_sale_b).ids)],
            'supplier_taxes_id': [(6, 0, (cls.tax_purchase_a + cls.tax_purchase_b).ids)],
        })
```

l10n_fr_ecotaxe_sale/tests/test_sale_ecotaxe.py
```diff
@@ -0,0 +1,124 @@
...
+        # I create a customer sale.
+        partner12 = self.env.ref("base.res_partner_12")
+        self.sale = self.create_sale_partner(
+            sale_amount=100.00, partner_id=partner12, product_id=self.product_b
+        )
+        # I assign a product with fixed ecotaxte to sale line
+        self.sale_line1 = self.sale.order_line[0]
+        self.sale_line1._compute_ecotaxe()
+        self.assertEqual(self.product_b.ecotaxe_amount, 16)
+        self.sale_line1.product_uom_qty = 3
```
ajaniszewska-dev, line 92: probably `self.sale_line1.product_uom_change()` missed here if needed at al?

l10n_fr_ecotaxe_sale/tests/test_sale_ecotaxe.py
```diff
@@ -0,0 +1,124 @@
...
+                "ecotaxe_product_status": "P",
+                "ecotaxe_supplier_status": "FAB",
+            }
+        )
+        cls.product_a.weight = 100
+        cls.product_a.ecotaxe_classification_id = cls.ecotaxe_classification1.id
+        cls.product_b.weight = 400
+        cls.product_b.ecotaxe_classification_id = cls.ecotaxe_classification2.id
+
+    def create_sale_partner(self, partner_id, product_id, sale_amount=0.00):
```
ajaniszewska-dev, line 40: i think sale_amount isn't needed to set price_unit on so line...but without that getting totally different vals on assertion . Can't find how this method looked like historically because it was removed.

test-requirements.txt
SilvioC2C: I agree with @ivantodorovich, please fix this

ajaniszewska-dev: @SilvioC2C  added message in PR description

ivantodorovich (approved):
Technically seems good 👍🏻 
I can only speak for that part as I don't know the functional case

test-requirements.txt
SilvioC2C: Sorry, my bad

ivantodorovich:
Btw, it'd be good to squash your "Fix tests", "Fix errors in view" commits into the "[MIG] Migration to 15.0 .. " one.
To keep history clean

SilvioC2C (approved):
LGTM! :+1:

l10n_fr_ecotaxe_sale/models/sale.py
```diff
@@ -0,0 +1,46 @@
...
+
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+
+    _inherit = "sale.order.line"
+
+    subtotal_ecotaxe = fields.Float(
+        store=True, compute="_compute_ecotaxe", oldname="amount_ecotaxe"
```
ivantodorovich:
```suggestion
        store=True, compute="_compute_ecotaxe"
```

Otherwise getting these warnings:
```
WARNING testdb odoo.fields: Field sale.order.line.subtotal_ecotaxe: unknown parameter 'oldname', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it 
```

`oldname` has been deprecated many versions ago, and in any case it should be removed during a migration IMO

ivantodorovich (changes requested):

ivantodorovich (approved):

PR #346, opened 2022-02-11
[14.0][FIX] l10n_fr_siret_lookup: zip is now given as int by the API

alexis-via:
'codedepartementetablissement' is not always set (in particular for companies in Corsica), so we now only use codepostaletablissement


alexis-via:
@remi-filament you may have noticed the breakage due to the change in the opendatasoft API. I fixed it here.

remi-filament:
thanks @alexis-via for fixing that !

PR #348, opened 2022-02-28
[FIX] company_registry : do not set company_registry field readonly on res.company model

legalsylvain:
Trivial patch.

the module ``l10n_fr_siret`` introduce a field ``company_registry`` field on ``res.partner`` and set the existing ``res.company`` field related to it. but the field is readonly on the company form.

this PR fixes this problem.

Note : this problem has been fixed upstream. see : https://github.com/OCA/l10n-france/blob/14.0/l10n_fr_siret/models/res_company.py#L22

CC : authors / contributors : @alexis-via, @fmdl, @zuher83, @remi-filament, @Kev-Roche 

GRAP : n°292 CC / @quentinDupont

### DIFF
--- a/l10n_fr_siret/models/company.py
+++ b/l10n_fr_siret/models/company.py
@@ -19,5 +19,8 @@ class ResCompany(models.Model):
         string='NIC', related='partner_id.nic', store=True, readonly=False)
     # company_registry field is definied in base module on res.company
     company_registry = fields.Char(
-        string='Company Registry', related='partner_id.company_registry',
-        store=True)
+        string='Company Registry',
+        related='partner_id.company_registry',
+        store=True,
+        readonly=False
+    )


remi-filament (approved):

legalsylvain:
@sbidoul : 
- the runboat fails because : ``account_facturx`` (/odoo/addons) module is incompatible with ``account_e-invoice_generate`` (OCA/edi)
See log : https://runboat.odoo-community.org/api/v1/builds/bdeef7f49-e9a3-4239-bc67-91af7ac04398/init-log
See exclude reference : https://github.com/OCA/edi/blob/12.0/account_e-invoice_generate/__manifest__.py#L15

In travis workflow, it is handled here https://github.com/OCA/l10n-france/blob/12.0/.travis.yml#L28

Do you know how to do the same thing on runboat ?

Thanks. CC : @OCA/local-france-maintainers

sbidoul:
> Do you know how to do the same thing on runboat ?

There is no way to configure runboat per repo. More precisely, the [foundation to configure per repo exists](https://github.com/sbidoul/runboat/blob/c5b2d215d74a86877b6cee3299738329b851473a/src/runboat/settings.py#L19), but it is part of the runboat configuration, and not exposed to repo maintainers.

Not sure what's best... we could may be run runboat for l10n-france on OCB which disables auto_isntall on a bunch of Odoo modules ? Or run runboat on OCB for all repos ? If possible I'd like to avoid ending up with a `.runboat.yml` in repos because that opens the door to a lot of complexity. And of course, parsing travis.yml or .github/workflow is a no no because it binds runboat to specific CI providers.

legalsylvain:
Thanks @sbidoul for your quick answer.
That's not trivial.

@alexis-via : I see that when you ported the edi module in V14, you removed the excludes part.
Another solution could be to remove it also in V12.

https://github.com/OCA/edi/commit/5bbe32c393ea634f5302b1b2f857ed71412ed7ac#diff-088676fd4df173919ef4de5a3e183389c83de6614fdfc2d5b0ef1cb56853c04aL15

What do you think ?

legalsylvain:
Proposal to fix CI available here : https://github.com/OCA/edi/pull/537

Kev-Roche:
Code Review, LGTM

PR #364, opened 2022-06-01
[12.0] Fix PyPDF2 version to <2.0

Kontrolix:
Fix PyPDF2 version to<2.0 since version 2 is incompatible with odoo 12.0


Kontrolix:
runboat check failed but I think it's not linked to my changes. Is somedody can confirm me that ?

ivantodorovich (approved):
Yeap, runboat is red but not related to this

PR #365, opened 2022-06-08
[15.0] Add some departments fr oversea

DilawBz:
We have added in this commit some departments of france oversea which were not included in your versions

### DIFF
--- a/l10n_fr_department_oversea/data/res_country_department.xml
+++ b/l10n_fr_department_oversea/data/res_country_department.xml
@@ -25,7 +25,7 @@
         <field name="state_id" ref="res_country_state_reunion" />
     </record>
     <!-- ### ADD AUNEOR-CONSEIL ### -->
-    <record 
+    <record
         id="res_country_department_sait_pierre_miquelon"
         model="res.country.department"
     >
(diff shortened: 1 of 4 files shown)


l10n_fr_department_oversea/data/res_country_department.xml
```diff
@@ -24,9 +24,57 @@
         <field name="code">974</field>
         <field name="state_id" ref="res_country_state_reunion" />
     </record>
+    <!-- ### ADD AUNEOR-CONSEIL ### -->
```
legalsylvain: Please remove the lines regarding auneor.

l10n_fr_department_oversea/data/res_country_department.xml
```diff
@@ -24,9 +24,57 @@
...
+        <field name="name">Polynésie Française</field>
+        <field name="code">987</field>
+        <field name="state_id" ref="res_country_state_polynesie_francaise" />
+    </record>
+    <record id="res_country_nouvelle_caledonie" model="res.country.department">
+        <field name="name">Nouvelle Calédonie</field>
+        <field name="code">988</field>
+        <field name="state_id" ref="res_country_state_nouvelle_caledonie" />
+    </record>
+    <record id="res_country_department_monaco" model="res.country.department">
```
legalsylvain: Monaco is a country.

legalsylvain (changes requested):
Thanks for jour contribution. Some blocking points inline.
Regards.

alexis-via:
I disagree with this PR ; so far, res.country.department was for real French departments (DOMs) and not TOMs. I think we should not change that.

alexis-via:
So I propose to close this PR.

PR #366, opened 2022-06-21
[14.0] [UPD] Update dotfiles from template and switch to Github Actions

ivantodorovich:
Pulling latest changes with `copier update`, and switching to Github Actions CI


ivantodorovich:
There's a relatively new check in OCA CI: `manifestoo check-licenses`.
It detects licenses inconsistencies in module dependencies.

Apparently we have one on this repo:

```
l10n_fr_fec_oca (LGPL-3) depends on date_range (AGPL-3)
```

@alexis-via : as module author, are you ok if we switch that module license to `AGPL-3` ?

alexis-via:
The module l10n_fr_fec_oca is a fork of l10n_fr_fec, which is LGPL with part of the copyright from Odoo SA. So we cannot change the licence. We could remove the dependency on date_range.

ivantodorovich:
I disabled the licenses check for now, so that we can move forward with this switch (and possibly unlock some blocked PRs)
I recommend we fix this inconsistency during a 15.0 migration. The `15.0` branch [already has this check enabled](https://github.com/OCA/l10n-france/blob/60a8f79c883f05e067624a4ca0b0783f4ca49da4/.github/workflows/test.yml#L58-L59) so we won't forget about it, the migration PR will be red if we don't solve it.

WDYT?

alexis-via:
Yes, ok for me.

alexis-via (approved):

PR #367, opened 2022-06-28
[14.0] [FIX] `l10n_fr_siret`: handle La Poste siret numbers (checksum validation rule exception)

ivantodorovich:
See:

- https://github.com/arthurdejong/python-stdnum/commit/73f5e3a864915e9f3f2b01683ab9b192b4217e46
- https://fr.wikipedia.org/wiki/Système_d'identification_du_répertoire_des_établissements#Exceptions_pour_le_groupe_La_Poste

This fix will be released in a future version of python-stdnum, but it's not in 1.17.

### DIFF
--- a/l10n_fr_siret/models/res_partner.py
+++ b/l10n_fr_siret/models/res_partner.py
@@ -1,20 +1,46 @@
-import logging
+from stdnum.fr.siren import is_valid as siren_is_valid
+from stdnum.fr.siret import is_valid as siret_is_valid
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
-
-logger = logging.getLogger(__name__)
-try:
-    from stdnum.fr.siren import is_valid as siren_is_valid
-    from stdnum.fr.siret import is_valid as siret_is_valid
-except ImportError:
-    logger.debug("Cannot import stdnum")
 
 
 class Partner(models.Model):
     """Add the French official company identity numbers SIREN, NIC and SIRET"""
 
     _inherit = "res.partner"
+
+    @api.model
+    def _validate_siren(self, number, raise_if_not_valid=False) -> bool:
+        valid = siren_is_valid(number)
+        if raise_if_not_valid and not valid:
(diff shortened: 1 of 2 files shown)


l10n_fr_siret/models/res_partner.py
```diff
@@ -1,21 +1,47 @@
-import logging
+from stdnum.fr.siren import is_valid as siren_is_valid
+from stdnum.fr.siret import is_valid as siret_is_valid
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
-logger = logging.getLogger(__name__)
-try:
```
ivantodorovich, line 7: FTR, encapsulating imports in `try` blocks is not longer necessary in recent Odoo versions.

Odoo won't import these `py` files unless the module is installed, and it can't be installed without the library: https://github.com/OCA/l10n-france/blob/c3208b5183dff5232988c376c2595f855b1751a8/l10n_fr_siret/__manifest__.py#L15

ivantodorovich:
Travis error seems unrelated.
Maybe we could fast-track https://github.com/OCA/l10n-france/pull/366 ?

alexis-via:
AFAIK, this PR is just to avoid waiting for python-stdnum 1.18. I think that we should push for a 1.18 release of python-stdnum instead of modifying the code in the module l10n_fr_siret. I asked the maintainer of python-stdnum for a new release, cf my last comment on https://github.com/arthurdejong/python-stdnum/issues/291

ivantodorovich:
> AFAIK, this PR is just to avoid waiting for python-stdnum 1.18. I think that we should push for a 1.18 release of python-stdnum instead of modifying the code in the module l10n_fr_siret. I asked the maintainer of python-stdnum for a new release, cf my last comment on https://github.com/arthurdejong/python-stdnum/issues/291

Indeed, that would be much better for sure !

alexis-via:
Stdnum 1.18 has been released on Nov 13 2022
https://pypi.org/project/python-stdnum/#history
So users just need to update their version of stdnum.

PR #368, opened 2022-07-11
[11.0] l10n_fr_intrastat_product: Add mig script

alexis-via:


ivantodorovich (approved):

legalsylvain (approved):
Thanks.

Note : I think the copyright of the files are wrong. (not a blocking point).

alexis-via:
It's a copy-paste from the module intrastat_product with only 1 change for the table name (l10n_fr_intrastat_product_declaration instead of intrastat_product_declaration), so I kept the original copyright.

PR #380, opened 2022-10-14
[14.0] l10n_fr_account_tax_unece: follow changes in l10n_fr

alexis-via:
Update module following a big change in l10n_fr
https://github.com/odoo/odoo/pull/84918

Several XMLIDs were changed in l10n_fr, which broke this module.


alexis-via:
The impact is bigger than I thought, because the tests of l10n_fr_account_vat_return are also broken. I disable those tests temporarily.

PR #381, opened 2022-10-19
[14.0] l10n_fr_account_vat_return: convert raise in chatter warning

alexis-via:
Don't block if account mapping is missing on fiscal position with fr_vat_type=france_exo because it may be a purchase-only fiscal position -> write warning in chatter.


legalsylvain:
FYI, you could also use ``web_notify`` OCA/module to display a more visible message.

PR #390, opened 2022-12-09
automatically remove payment_mode_id on refund

Kontrolix:
Since payment_mode_id must be empty on refund with transmit_method_code == "fr-chorus", if we try to create a refund with the account.invoice.refund wizard and filter_refund set to cancel or modify that auto validate the refund we get the error ''The Payment Mode must be empty for customer refunds sent to Chorus." I suggest to remove it in _prepare_refund to avoid to get this error and allow us to use the wizard normally.


alexis-via:
Did you test your PR ?
There is no "return values" at the end of the method, so I don't see how it could possibly work :-(

Kontrolix:
Sorry for this gross error... I use this code in one of my module where I overide other fields too. I made just an extract here with only the payment_method_id field, and I forgot the return, so yes I have to admit it, i didn't exactly test the exact code of the PR, my bad.

PR #392, opened 2022-12-09
[16.0][MIG] l10n_fr_siret_lookup

alexis-via:


alexis-via:
Requires https://github.com/OCA/l10n-france/pull/391

PR #394, opened 2022-12-13
[15.0][IMP] l10n_fr_department: improve performances

SilvioC2C:
On install, ``_compute_department()`` method is triggered by the `post-init-hook` on every existing French partners, archived ones as well. This commit tries to improve performances by grouping partners by zipcode (if valid) to avoid huge loading times when installing the module.

Tested on a DB where recomputation was triggered on >75k records by the ``post-init-hook``: install time dropped from >60mins to ~2.5mins.

Edit: down to ~2.0mins after @ivantodorovich suggestions.

### DIFF
--- a/l10n_fr_department/model/res_partner.py
+++ b/l10n_fr_department/model/res_partner.py
@@ -3,6 +3,8 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
+from odoo.tools.cache import ormcache
+from odoo.tools.misc import groupby
 
 
 class ResPartner(models.Model):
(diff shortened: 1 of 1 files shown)


yankinmax (approved):
@SilvioC2C interesting implementation, thanks for the work done.

l10n_fr_department/model/res_partner.py
```diff
@@ -18,33 +20,35 @@ class ResPartner(models.Model):
...
     # If a department code changes, it will have to be manually recomputed
     def _compute_department(self):
-        rcdo = self.env["res.country.department"]
+        def _get_zipcode(p) -> str:
+            if not p.country_id.id or p.country_id.id not in fr_country_ids:
+                return ""
+            p_zip = p.zip
+            if not p_zip or len(p_zip) != 5:
+                return ""
+            return p_zip.strip().replace(" ", "").rjust(5, "0")
```
ivantodorovich: Seems weird to check for the len to be exactly `5` before the normalization done by  strip/rjust.
It would give false `""` in case of `p_zip = "    12345"`, no?

How about something like this instead?

```suggestion
            if not p_zip:
                return ""
            p_zip = p_zip.strip().replace(" ", "").rjust(5, "0")
            if not len(p_zip) != 5:
                return ""
            return p_zip
```

l10n_fr_department/model/res_partner.py
```diff
@@ -18,33 +20,35 @@ class ResPartner(models.Model):
     @api.depends("zip", "country_id", "country_id.code")
     # If a department code changes, it will have to be manually recomputed
     def _compute_department(self):
-        rcdo = self.env["res.country.department"]
+        def _get_zipcode(p) -> str:
```
ivantodorovich:
```suggestion
        def _get_zipcode(partner) -> str:
```

On this method, I recommend renaming `p` to `partner` and `p_zip` to `partner_zip`.
The abbreviation seems unnecessary and makes it more difficult to read IMO

l10n_fr_department/model/res_partner.py
```diff
@@ -18,33 +20,35 @@ class ResPartner(models.Model):
...
+                    self.env["res.country.department"]
+                    .search(
+                        [("code", "=", code), ("country_id", "in", fr_country_ids)],
+                        limit=1,
+                    )
+                    .id
                 )
-                dpt_id = dpt and dpt.id or False
-            partner.department_id = dpt_id
+            self.browse().concat(*partner_list).department_id = dpt_id
```
ivantodorovich: If you wanna be even more efficient, you could move the search out of the loop.
Something like this:

```py
        partners_by_zip = groupby(self, key=_get_zipcode)
        department_codes = [
            self._fr_zipcode_to_department_code(zipcode)
            for zipcode in partners_by_zip.keys()
            if zipcode
        ]
        departments = (
            self.env["res.country.department"].search(
                [("code", "in", department_codes), ("country_id", "in", fr_country_ids)]
            )
            if department_codes
            else self.env["res.country.department"]
        )
        departments_by_code = groupby(departments, key=lambda d: d.code)
        for zipcode, partner_list in groupby(self, key=_get_zipcode):
            department = self.env["res.country.department"]
            if zipcode:
                code = self._fr_zipcode_to_department_code(zipcode)
                department = departments_by_code[code][0]
            self.browse().concat(*partner_list).department_id = department
```

Note that `_fr_zipcode_to_department_code` is called twice per zipcode, however thanks to the `ormcache` the second time will be super quick.

However, we could group the partners by department code instead of grouping by zip, and so we wouldn't need to call this method twice

ivantodorovich (approved):
Nice work!
Non-blocking suggestions

l10n_fr_department/model/res_partner.py
SilvioC2C: @ivantodorovich seems weird to me too, but I wanted to keep this part as it was originally, without changing the business logic too much. Not sure what the initial idea was, so I don't feel confident in changing that...

SilvioC2C:
@ivantodorovich I've updated the PR with your precious suggestions, can you please review it once more? :sweat_smile:

ivantodorovich (approved):
Looks good ;)

PR #397, opened 2022-12-19
[16][MIG] account_statement_import_fr_cfonb

florian-dacosta:
Standard Migration + minor improvement about complementary info.
depends on https://github.com/OCA/bank-statement-import/pull/535

@alexis-via


alexis-via:
I had some difficulty to find info on LCC, because it is not in the official CFONB spec (only LIB and MMO are in the official spec). I eventually found it in slide 21 of this sage presentation: https://slideplayer.fr/slide/5382777/
LCC = Libellé de Client à Client

alexis-via:
I think if we add LCC, we should also add LC2 and the other flags

alexis-via:
Works with the new PR https://github.com/OCA/bank-statement-import/pull/575

PR #398, opened 2022-12-26
[16.0][MIG] l10n_fr_state

alexis-via:
Re-write pre-init hook to mutualize the code with l10n_fr_department_oversea and avoid hard-coding XMLIDs in the code


legalsylvain (approved):
Thanks !

Issue #384, opened 2022-11-22
[14.0]l10n_fr_department conflict with res_partner_department


flotho:
trying to install https://github.com/OCA/partner-contact/blob/14.0/partner_contact_department/models/res_partner.py#L11 and l10n_fr_department in the same db cause issues.

what is the process to deal with this when 2 fileds with the same name are not referencing the same comodel?

legalsylvain:
There is no way. For the time being, both module are incompatible.
One solution could be to rename the field in the current module to another one.
But you have to add migration script. quite a mess.

flotho:
as a starting point, is there any options in the manifest that could prevent the double installation ?

alexis-via:
For v14, I made a PR to declare a conflict in the manifest.
For v16, we could change the name of the field "department_id" on res.partner. What about "country_department_id" ? Feel free to propose a better field name.

PR #399, opened 2022-12-26
[16.0][MIG] l10n_fr_department and l10n_fr_department_oversea

alexis-via:
fix bug https://github.com/OCA/l10n-france/issues/384 : field department_id on res.partner has been renamed to country_department_id


alexis-via:
Requires https://github.com/OCA/l10n-france/pull/398

legalsylvain (changes requested):
Thanks for porting this module. 

Migration script is missing.

Otherwise, LGTM.

alexis-via:
As it is a computed field, a migration script is not required, but it could save same time during migration.

l10n_fr_department/model/res_partner.py
```diff
@@ -10,16 +10,19 @@
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
-    department_id = fields.Many2one(
+    # In v16, this field was renamed from department_id to country_department_id
+    # to avoid a field name conflict with the module partner_contact_department
+    # from https://github.com/OCA/partner-contact
```
ivantodorovich: These comments belong to the commit message or to the PR itself.

They're only useful when reviewing/analyzing changes from previous code.
But they have no use at all in the present/future code

l10n_fr_department/view/res_country_department.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <!--
-  Copyright 2013-2020 GRAP (http://www.grap.coop)
+  Copyright 2013-2022 GRAP (http://www.grap.coop)
```
ivantodorovich, line 3: FTR you could save yourself from ever having to update these Copyright lines.
When reading the copyright comment, the dates are to be read as "starting.." (and legally it'll be valid for up to 70 years from that point)

eg: `Copyright 2013 GRAP (http://www.grap.coop)`

Means it's copyright is valid from 2013 to 2083.
There's no need at all to put an end date, that end up being just noisy diffs, and it's totally useless.


https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

legalsylvain:
> As it is a computed field, a migration script is not required, but it could save same time during migration.

Indeed. But a little weird to raise the recompute of an existing field. And on res.partner table, it can take time. Better to use rename_fields of openupgradelib.

alexis-via:
I added and tested the migration script. Requires a recent version of openupgradelib to support the new system for translations.

legalsylvain (approved):
Thanks !

PR #406, opened 2023-01-04
[13.0][MIG] l10n_fr_department_delivery

BenjaHe:


alexis-via (approved):
code review

BenjaHe:
hello @alexis-via. Thanks for the code review.
This PR has a problem with Travis, but I do not understand why. Do you have any idea ?

alexis-via:
Why do you inherit the wizard and not just the method _match_address() on delivery.carrier ? It seems to me that the method _match_address() is specifically designed to be inherited in this kind of cases.

alexis-via:
I migrated this module to v14 using the inherit of the method _match_address() on delivery.carrier and it works well.
https://github.com/OCA/l10n-france/pull/443

PR #417, opened 2023-02-15
[15.0][IMP] l10n_fr_intrastat_product: We don't need fiscal representative if partner is from Europe

yankinmax:
It seems we can have a note with the fiscal representative even if partner and invoice are in the Europe.


yankinmax:
@alexis-via @gurneyalex 
also pls consider this one which updates pre-commit with latest changes by copier:
- [ ] https://github.com/OCA/l10n-france/pull/418

damdam-s (approved):

PR #421, opened 2023-03-04
[14.0][FIX] l10n_fr_account_vat_return: fix reconciliation

alexis-via:
Before reconciliation, we need to remove partial reconcile otherwise we have an error message which says that the moves are already reconciled.


alexis-via:
GRRRR, pre-commit error is unrelated

alexis-via:
All green! Thank you @legalsylvain

PR #422, opened 2023-03-04
[REF] copier update v1.14.2

legalsylvain:
Hi @alexis-via.

This should CI problem mentionned here :  https://github.com/OCA/l10n-france/pull/421

```
git checkout 14.0
git checkout -b 14.0-fix-copier
copier update
git push your_remote 14.0-fix-copier
```

Generaly, if CI breaks, you can check issues / PR on this project. https://github.com/OCA/oca-addons-repo-template/

Here is the bug in question : https://github.com/OCA/oca-addons-repo-template/pull/184

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.14.1
+_commit: v1.14.2
 _src_path: gh:oca/oca-addons-repo-template
 ci: GitHub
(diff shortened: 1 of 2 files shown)


alexis-via:
Thank you so much Sylvain !

legalsylvain:
you're welcome. 

I find it a bit complicated to understand how the OCA CI works (and how to fix it if there are problems), but once you get the hang of it, it works great and the https://github.com/OCA/oca-addons-repo-template/ project maintenance team is super responsive

PR #428, opened 2023-03-20
[14.0][FIX] l10n_fr_account_vat_return: only consider revenue account mapping for exo operations

alexis-via:


acalmett (approved):
Using this module in production. No code review.

PR #431, opened 2023-03-27
[16.0][IMP] account_statement_import_fr_cfonb: add small hook to allow to a…

florian-dacosta:
…dd more info in label

@alexis-via 
As you know, I have already added back some complementary label in the module, but it is still not enough.
I have BNP Cfonb files that add the sale order reference in the RCN category (in case of customer bank transfer payment)
In other cases, the information is useless indeed, but it is impossible not to have it in some other case...

So, in order to not pollute people that don't need it, I propose this small hook that allow to add some complementary code at demand.
What do you think ?

### DIFF
--- a/account_statement_import_fr_cfonb/wizard/account_statement_import.py
+++ b/account_statement_import_fr_cfonb/wizard/account_statement_import.py
@@ -18,6 +18,10 @@ class AccountStatementImport(models.TransientModel):
 
     _excluded_accounts = []
 
+    @api.model
+    def _get_allow_cfonb_complementary_types(self):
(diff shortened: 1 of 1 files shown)


florian-dacosta:
@alexis-via As we talked about some time ago, I've added the RCN complentary code by default as it is used a lot by the bank and sometimes with really important information (like SO number in case of wire transfer)
Can you have a quick look on the PR?

florian-dacosta:
@alexis-via Friendly reminder, could you make a review on this trivial PR?

florian-dacosta:
@alexis-via

alexis-via (approved):

PR #439, opened 2023-04-16
[FIX] l10n_fr_department: fast install

fmdl:
Before this commit with a large DB it can take some hours, after a few minutes.

@alexis-via


legalsylvain (commented):
Hi @fmdl. Thanks for your patch. 
1) I wonder if it's better to improve the algorighm of ``_compute_department`` directly. so, the improvment will be available everywhere. (in case of mass editing, in case of mass import, etc..)
2) could you fix pre-commit error  ?

thanks.

PR #440, opened 2023-04-17
[16.0][MIG] l10n_fr_account_vat_return and l10n_fr_account_vat_return_teledec

alexis-via:
There is 1 pb left: migration of the analytic account to analytic distribution on l10n.fr.account.vat.box. It don't know how to handle that because there is a company_dependent=True on that field!
My ideas:
1) either I create a new object with a O2M from l10n.fr.account.vat.box that give the value of the distribution per company
2) either I consider that the analytic account should be set by the distribution model, but it's a bit wired and not intuitive


alexis-via:
FYI, this PRwill be cancelled and replaced by a forward port of this other PR https://github.com/OCA/l10n-france/pull/470

alexis-via:
Replaced by https://github.com/OCA/l10n-france/pull/475

PR #445, opened 2023-05-05
[IMP] l10n_fr_department : if country is not defined on partner, use the country of the company of the partner, to guess the department

legalsylvain:
@alexis-via : could you take a look on this trivial one ?

context : a lot of partner has no country defined. Some department computation is not done. However, we can assume that the country is the country of the current company of the partner.

Thanks !


legalsylvain:
Used in production since 6 monthes.

/ocabot merge patch

PR #466, opened 2023-06-30
[16.0] [IMP] l10n_fr_department : Change label of country_department_id to fix warning when installed with hr module

myrrkel:
On odoo.sh when installed with hr module, we have this warning :
"Two fields (country_department_id, department_id) of res.users() have the same label: Department. [Modules: None and hr]"


legalsylvain (commented):
I guess you should also change the pot file.
https://github.com/HodeiBot/l10n-france/blob/16.0/l10n_fr_department/i18n/l10n_fr_department.pot#L52
don't you think ?

legalsylvain (approved):
/ocabot merge patch

PR #469, opened 2023-07-04
[16.0][ADD] l10n_fr_pos_caisse_ap_ip

alexis-via:
New module l10n_fr_pos_caisse_ap_ip that adds support for the Caisse AP protocol over IP of the association du paiement: 

https://www.associationdupaiement.fr/protocoles/protocole-caisse/

This module is a major step forward compared to pos_payment_terminal:

* it doesn't need POSbox nor pywebdriver
* it implements bi-directional communication: the answer of the payment terminal is parsed and, in case of success, the payment line (if partial payment) or the full order is automatically validated.

But the odoo server needs to be able to reach the payment terminal.

As this protocol is specific to France, we decided to propose it here, to avoid confusion to users (I have too many emails from users in other countries trying desperately to use the solution outside France !)


l10n_fr_pos_caisse_ap_ip/static/src/js/payment_caisse_ap_ip.js
```diff
@@ -0,0 +1,105 @@
+/*
+    Copyright 2023 Akretion France (http://www.akretion.com/)
+    @author: Alexis de Lattre <alexis.delattre@akretion.com>
```
fclementic2c, line 3: could you add camptocamp as financial contributor please ? thanks @alexis-via

legalsylvain: The place should not be in the header of the file. (but can be set in a CREDITS.rst file, in the readme section).

PR #470, opened 2023-07-14
[14.0] l10n_fr_account_vat_return: fix for DGFiP consistency checks, improve negative box support, add support for imports (extracom products)

alexis-via:
Easier inheritance of tax search domains
Filter out the non-VAT taxes using UNECE type code

Next step : adjust rounding to have a sum of due vat base = some of taxed operations


alexis-via:
adjust rounding to have a sum of due vat base = some of taxed operations: DONE
I also switch from PyPDF2 (unmaintained) to pypdf (actively maintained)

alexis-via:
FYI, I plan to merge it rapidly now, to have the fix for DGFiP consistency checks for the VAT return of August

PR #473, opened 2023-09-05
Update company.py

BriceCG:
Change Payload from info to debug as they are huge and not needed in production


klodr (approved):
👍

legalsylvain (commented):
Thanks ! 

/ocabot merge patch

PR #474, opened 2023-09-06
Update company.py

BriceCG:
Change Payload from info to debug as they are huge and not needed in production


klodr (changes requested):
Nope this one remains !

legalsylvain (approved):
/ocabot merge patch

PR #482, opened 2023-09-26
[ADD] new module l10n_fr_pos_cert_update_draft_order_line.

legalsylvain:
This module fixes the Odoo module 'French Point of Sale Certification'
to allow cashier to update order lines, when order is **not** confirmed.

* The law stipulates that the tool must not be used to modify confirmed orders,
  but does not specify that it is forbidden to change a quantity if the order
  is in the process of being seized.

* If the cashier make a mistake, the current Odoo design generates hugly and
  incomprehensible tickets for the customer.

* In fact, Odoo core module is totally inconsistent in its understanding of the law,
  since it prohibits the modification of a line in a draft order,
  but authorizes the deletion of the full draft order.


gregory-moka (approved):
Fonctionnal review > OK

LESTRAT21:
Agree with you @legalsylvain  concerning the unconsistant rule of Odoo module that enables to delete a ticket but prohibits the modification of a line in a draft order!
I did a functional review and installed your new module l10n_fr_pos_cert_update_draft_order_line
Thanks for this module !

legalsylvain:
Hi @LESTRAT21 thanks for you review.
Just a point regarding validation of PR : 
You can go on "File changed" tab, then click on "Review Changes", click on "Approve" and let a comment.
That way, bots knows that the PR is validated. (and Repo Responsible can merge it).

LESTRAT21 (approved):
Agree with you @legalsylvain concerning the unconsistant rule of Odoo module that enables to delete a ticket but prohibits the modification of a line in a draft order!
I did a functional review and installed your new module l10n_fr_pos_cert_update_draft_order_line
Thanks for this module !

PR #490, opened 2023-11-15
[16.0] Add module l10n_fr_account_banking_pain_base

alexis-via:
This module adds the regulatory reporting codes used in France for the generation of ISO 20022/PAIN credit transfer and direct debit files. These regulatory codes are required for non-SEPA credit transfers with an amount over 50 000 €.


alexis-via:
Requires https://github.com/OCA/bank-payment/pull/1174

PR #500, opened 2023-12-20
17.0 mig l10n fr siret

youelEl:


l10n_fr_siret/views/res_partner.xml
```diff
@@ -1,67 +1,50 @@
-<?xml version="1.0" encoding="utf-8" ?>
+<?xml version="1.0" encoding="utf-8"?>
 <odoo>
 
     <!-- Add the SIREN and RC -->
     <record model="ir.ui.view" id="res_partner_form_l10n_fr">
         <field name="name">res.partner.form.siret</field>
         <field name="model">res.partner</field>
-        <field name="inherit_id" ref="l10n_fr.res_partner_form_l10n_fr" />
+        <field name="inherit_id" ref="base.view_partner_form"/>
```
alexis-via, line 8: You should not change that! The siret field is added by l10n_fr, so, if you inherit the field, you need to inherit the proper view

alexis-via:
Replaced by #502

PR #507, opened 2024-01-11
[16.0][ADD] l10n_fr_department_product_origin

legalsylvain:
based on the pending PR : https://github.com/OCA/product-attribute/pull/1389
(+ changes done in https://github.com/acsone/product-attribute/pull/8)

### Description

This module extends the functionality of Product Origin module to add french department on products.

product_form

CC : @quentinDupont, @alexis-via, @acsonefho


test-requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/OCA/product-attribute@refs/pull/1389/head#subdirectory=setup/product_origin
```
legalsylvain: Hi @sbidoul. 

Question : once setup folder has been removed, replaced by toml file, what is the correct syntax to write such dependency in the futur ? 

thanks !

sbidoul: @legalsylvain `subdirectory` is the directory where the python project is. The python project is where the setup.py or pyproject.toml is. So, just replace `setup/product-origin` with `product_origin`.

See also https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

legalsylvain: Thanks ! very clear.

PR #513, opened 2024-01-23
[14.0] l10n_fr_account_vat_return: millesime 2024 + several fixes and improvements

alexis-via:
Add support for hr_expense scenario in autoliquidation
For untaxed operations, compute period balance only in sale journal, to exclude cutoff entries


alexis-via:
pre-commit error was completely unrelated, so I had to merge manually

PR #515, opened 2024-01-31
[FIX] l10n_fr_siret[_lookup]: remove wrong dependency

yajo:
Odoo 16 depends on both requests and python-stdnum, so there's no need to specify those dependencies.

Besides, [python-stdnum is pinned to 1.16][1]. Thus, when applying https://github.com/OCA/oca-ci/pull/66 locally, the constraint failed. OCA addons should work with the dependencies pinned upstream.

[1]: https://github.com/odoo/odoo/blob/fe4d8e8014e50d4335e0a8b052b5098f3a79bcec/requirements.txt#L48

@moduon MT-4520 @EmilioPascual


l10n_fr_siret_lookup/__manifest__.py
```diff
@@ -14,7 +14,6 @@
     "depends": [
         "l10n_fr_siret",
     ],
-    "external_dependencies": {"python": ["requests", "python-stdnum>=1.18"]},
```
sbidoul, line 17: It's okay to specify direct dependencies here, if only for docs purposes. So the line must not be removed.

Maybe there a good reason to require `python-stdnum>=1.18` for this module to work? In such case we must accept it.

yajo: > So the line must not be removed.

Restored

l10n_fr_siret/__manifest__.py
```diff
@@ -12,7 +12,7 @@
     "website": "https://github.com/OCA/l10n-france",
     "license": "AGPL-3",
     "depends": ["l10n_fr", "base_view_inheritance_extension"],
-    "external_dependencies": {"python": ["python-stdnum>=1.18"]},
+    "external_dependencies": {"python": ["python-stdnum"]},
```
remi-filament, line 15: You should keep it here also
```suggestion
    "external_dependencies": {"python": ["python-stdnum>=1.18"]},
```

l10n_fr_siret_lookup/__manifest__.py
remi-filament: Hi @yajo thank you for looking into this,
as a matter of fact there is a good reason to request python-stdnum >= 1.18, which is this commit : https://github.com/arthurdejong/python-stdnum/commit/73f5e3a864915e9f3f2b01683ab9b192b4217e46

The module should work with older version, although it will fail for the specific SIRET covered by the above commit (but this specific case is not tested in the module so unit tests should not fail neither if done with older version)

remi-filament (changes requested):

alexis-via:
Yes, it's very annoying that Odoo depends on old version of python-stdnum. It's a lack of respect for the excellent work of the guys of python-stdnum. By depending on such an old version, we miss all the bugfixes of the recent years.

sbidoul:
I would not merge this. We'll find a better solution for https://github.com/OCA/oca-ci/pull/66

yajo:
What about this?
1. Add to the installation instructions that it is recommended to upgrade this library.
2. Leave it unconstrained.
3. Open PR to Odoo explaining the problem.

alexis-via:
@yajo I'm ok to remove the constraint on the version of python-stdnum in the manifest and explain in the README of the module that we recommend to have a version >= 1.18 to have the fix on SIRET that is included in that version.

yajo:
seems fair, done.

yajo:
Thanks, bot. AFAIK this is ready to merge. I just rebased to fix conflicts.

PR #522, opened 2024-02-26
[14.0][FIX] l10n_fr_account_vat_return: nice error message instead of crash in case of bad setup

alexis-via:
Backport of https://github.com/OCA/l10n-france/pull/521


alexis-via:
The pre-commit errors are unrelated, so I had to merge manually.
If someone knows how to fix these pre-commit errors which seems related to python 3.12, please come and help !

PR #523, opened 2024-03-01
[16.0][IMP] vat_return: warn if draft moves are present (instead of raising an error)

alexis-via:
When draft moves where present before end_date of the VAT return, we used to raise an error. With this commit, we continue to raise an error if option 'fr_vat_update_lock_dates' is enabled on the accounting config page or if one of the draft moves is the move of the previous VAT return. Otherwise, we start a wizard that shows the draft moves and gives the choice to the user between processing the draft moves or ignoring them.


alexis-via:
draft_move_v2

syera94 (approved):

PR #547, opened 2024-05-13
[IMP] account_banking_fr_lcr: add posibility to disable CRLF within generated CFONB files

AnizR:
Small improvement of current module:

- Adds a boolean field on payment method that allows to disable/enable addition of CRLF within generated CFONB files. By default, it is set to 'True' so, there isn't any modification in the behavior of the module.


acsonefho (approved):
GJ :muscle:

dreispt (approved):

alexis-via:
You had the issue with which bank ?

AnizR:
> You had the issue with which bank ?

Hello, it was for the bank 'LCL' (Le Crédit Lyonnais)

Issue #501, opened 2023-12-29
l10n_fr_siret: company_registry field on res.partner is native since v16


alexis-via:
The stupid guy who migrated l10n_fr_siret to v16 (me !) didn't spot that the field company_registry was added by Odoo SA in the **base** module, cf https://github.com/odoo/odoo/blob/16.0/odoo/addons/base/models/res_partner.py#L225
This field is used to store the national company identification number, cf for example the code of l10n_be and l10n_ro (they auto-extract it from the VAT number).

From my understanding, the field company_registry of l10n_fr was designed to write the name of the RCS where the company was registered. For example, for Akretion France which is RCS Lyon 792377731, then company_registry would store "Lyon". I don't use that field myself, and I don't find it useful. l10n_fr_siret was ported to v16 one year ago, and nobody found the bug yet... so I guess that I'm not the only one who don't use that field !!!

2 options:
1) remove company registry in l10n_fr_siret because we think that it's not useful
2) rename the field: for example **fr_company_registry**

I vote for option 1.

legalsylvain:
Thanks for your investigation !
Option 1.

matthieusaison:
🗳️ Option 1

hparfr:
Option 1

zuher83:
Option 1

PR #549, opened 2024-05-26
[16.0][FIX] Don't re-define field company_registry

alexis-via:
Fix bug #501: don't redefine field company_registry on res.partner ; that field is defined in the "base" module since odoo v16


rvalyi (approved):

legalsylvain:
thanks.

/ocabot merge minor

Issue #483, opened 2023-10-18
Generalization/Internationalization of l10n_fr_ecotaxe


dcorio:
Hi! Probably not the right place to ask about this...

Here in Italy we manage some excise duties in the very same way you manage ecotaxes.
I believe the same happens in other countries where excises are used to tax the import/export of special products like alcohol, batteries, petroleum derivatives or in general taxes base on weight, volume or other attributes.

Should we generalize the way OCA manage such taxes so each l10n-* project would share part of the code?

Cheers!

legalsylvain:
CC : @alexis-via could you take a look on this one ? 
thanks !

bealdav:
It seems to be a nice idea @mourad-ehm @alexis-via

bealdav:
Maybe we could put ecotax module here https://github.com/OCA/account-fiscal-rule

bealdav:
We now use account.tax for ecotax

https://github.com/OCA/account-fiscal-rule/pull/428 
https://github.com/OCA/account-fiscal-rule/pull/429

It's wip.

Could you have a quick look into.

cc @legalsylvain @dcorio

legalsylvain:
Hi @bealdav. No need to ping me on that topic. I don't use ecotax. (and really don't know how it works !) 
This one is for @akretion team !

rvalyi:
the work is largely done already in thr OCA/account-fiscal-rules repo. You can follow it here https://github.com/OCA/account-fiscal-rule/issues/453

I think this one can be closed meanwhile.

PR #553, opened 2024-06-05
[17.0] [REM] remove l10n_fr_ecotaxe

gurneyalex:
This should not have been merged, the modules are being generalized and moved to https://github.com/OCA/account-fiscal-rule

See #483


hparfr (approved):

gurneyalex:
@sbidoul Is this ok or will it have nasty side effects on pypi / weblate / others ?

Similar question for #552 (same PR on 16.0)

sbidoul:
@gurneyalex removing modules is fine. They will stay on PyPI though. If we want to discourage folks to install them, we have the option to yank the releases on PyPI.

sbidoul:
Ping me when it's merged, I'll check weblate as I'm not sure it handles removals.

gurneyalex:
@sbidoul both PRs merged

PR #555, opened 2024-06-07
[17.0][WIP] account_factoring_receivable_balance

dreispt:
Port from https://github.com/akretion/odoo-factoring/tree/16.0 and migration to 17.0
Related to https://github.com/akretion/odoo-factoring/issues/20


account_factoring_receivable_balance/models/subrogation_receipt.py
```diff
@@ -0,0 +1,265 @@
...
+            domain + [("move_id.currency_id", "=", journal.currency_id.id)]
+        )
+        lines.write({"subrogation_id": self.id})
+        vals = {"item_ids": [(6, 0, lines.ids)]}
+        if not self.statement_date:
+            statement = self.env["account.bank.statement"].search(
+                [
+                    ("journal_id.factor_type", "=", self.factor_type),
+                    ("journal_id.currency_id", "=", self.currency_id.id),
+                    ("state", "=", "confirm"),
```
max3903: @dreispt `account.bank.statement` has no `state` in version 17.0.

max3903:
@Murtaza-OSI @dreispt I was able to get move lines from invoices and generate the txt file with 101 lines.

I didn't get the move lines from credit notes, so no 102 lines.

Murtaza-SerpentCS:
> @Murtaza-OSI @dreispt I was able to get move lines from invoices and generate the txt file with 101 lines.
> 
> I didn't get the move lines from credit notes, so no 102 lines.

@max3903 file have 101 and 102 lines in .TXT file 
image

max3903:
@Murtaza-SerpentCS do you get those lines in the form view after clicking on "Calculate"?

Murtaza-OSI:
> @Murtaza-SerpentCS do you get those lines in the form view after clicking on "Calculate"?

Yes getting line 
fact

max3903:
@Murtaza-OSI Smart button is fixed and lines are correct in `line_ids`:

Screenshot-from-2024-06-13-20-02-41

CC: @dreispt

bealdav:
Thanks for porting this module. Nice Improvements

I asked myself if this module, is not generic enough to move an account OCA repository ?

Happy to hear there

bealdav:
@dreispt as I proposed `account_factoring_receivable_balance` can move to a non localization repo and other modules could stay here . What do you think ?

But it can be done at the end of your process here

dreispt:
@bealdav I agree in principle, but in the immediate term it is easier to work and refactor if they are in the same repo.
Once this is stabilized and before a merge, we can introduce those changes.

account_factoring_receivable_balance/models/subrogation_receipt.py
```diff
@@ -0,0 +1,237 @@
...
+    _inherit = ["mail.thread", "mail.activity.mixin"]
+    _check_company_auto = True
+    _rec_name = "factor_type"
+    _order = "target_date DESC"
+    _description = "Customer balance data for factoring"
+
+    factor_journal_id = fields.Many2one(
+        comodel_name="account.journal",
+        string="Journal",
+        domain="[('factor_type', '!=', False)," "('type', '=', 'bank')]",
```
bealdav, line 20: @dreispt 
In my experience (minor) but the journal associated to account.move used to send data to factor etablishment (invoices and credit notes) can be a `misc` one.
In this kind of journal you have not a lot settings, then it's less confusing to use this kind in a such context.
At least, we may consider to offer a choice.

A bank type is require to produce moves with statements data, that's not the same flow

Don't you think Daniel ?

dreispt: @bealdav Good question, this was something I struggled with.
For FactoFrance we need to also report the payments on Invoices.
So I settled with using a Bank journal, allowing you to post payments on factored invoices.

account_factoring_receivable_balance/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import fields, models
+
+
+class Partner(models.Model):
+    _inherit = "res.partner"
+
+    factor_journal_id = fields.Many2one(
+        comodel_name="account.journal",
+        domain="[('factor_type', '!=', False)]",
```
bealdav, line 12: Required in multi-company environment
```suggestion
        domain="[('factor_type', '!=', False)]",
        company_dependent=True,
```

account_factoring_receivable_balance_factofrance/models/account_journal.py
```diff
@@ -0,0 +1,15 @@
...
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    factor_type = fields.Selection(
+        selection_add=[("factofrance", "FactoFrance")],
+        ondelete={"factofrance": "set null"},
+    )
+    partner_identification_type = fields.Selection(
+        selection=[("1", "1 SIRET"), ("4", "4 TVA")],
+    )
```
bealdav, line 15: @dreispt  FYI in my repo I have made this settings to match with arbitrary factor keys.
image

https://github.com/akretion/odoo-factoring/pull/21/files#diff-2bf0078cef96b202cab38f4dcae616055927c9c3e252cc6f6b1f74a280737c02R18-R39

May be it could be generic ?

dreispt: IMO that is a bit complicated.
I would go for the simpler solution, and add those additional settings as fields, only visible for tis Factor Type.
Note that the "Factor Code" is really a "Factor Contract Number", so renaming it could make sense.

bealdav: It depends on your factor settings, in your case you only have one key but like any carrier dev you may have misc keys with unpredictable name or mean unshareable. If you have several factors for different companies in your ERP it becomes crazy.
But yes,  it could be not so generic.

dreispt: Well, the DB will have all those fields, but the user form would show only the relevant ones for the selected Factor.
Like what the shipping connectors do, as you pointed out.

bealdav: A matter of taste

account_factoring_receivable_balance/models/subrogation_receipt.py
rvalyi: > @bealdav Good question, this was something I struggled with.
> For FactoFrance we need to also report the payments on Invoices.
> So I settled with using a Bank journal, allowing you to post payments on factored invoices.

@dreis for Factofrance I strongly recommend you look at this other module I did initially, it is used in production since 3 years in a very big company and was designed for FactoFrance: https://github.com/akretion/odoo-factoring/tree/14.0/account_factoring

You can find a demo here: https://youtu.be/N7x7MtfEF1g?si=0FbU3P1O8Iv8MgzE

dreispt:
Rebased on recent 17.0 and added multi-company support.

PR #560, opened 2024-06-15
[BIG][16.0] account_banking_fr_lcr: add support for paper LCR, billet à ordre, Dailly, escompte, etc...

alexis-via:
So far, the module only supported LCR directes. This commit adds support for LCR acceptées and billets à ordres.


alexis-via:
Requires https://github.com/OCA/bank-payment/pull/1174

alexis-via:
Add button on customer invoices with payment mode code = 'fr_lcr' and fr_lcr_type = 'accepted' to get the paper letter of change as PDF.

 Add support for Dailly convention and cash discounts (escompte) with value date (date de valeur) on payment order (with default values on payment mode).

Add field on customer invoice to explicitely select the bank account of the customer that will be debited.

Improve CFONB file generation to be more compliant with the specs, in particular on references.

Add SIREN of company and customer in CFONB file if l10n_fr_siret module is installed and the SIREN field is set on the partner.

alexis-via:
lcr_short

PR #561, opened 2024-07-18
[IMP] l10n_fr_siret: warning for identical SIRET

vvrossem:
adds a warning banner on the partner form view either
- if another partner has the same SIRET,
- or if another partner has the same SIREN but different SIRET.

### DIFF
--- a/l10n_fr_siret/models/res_partner.py
+++ b/l10n_fr_siret/models/res_partner.py
@@ -56,8 +56,14 @@ class Partner(models.Model):
     )
     same_siren_partner_id = fields.Many2one(
         "res.partner",
-        compute="_compute_same_siren_partner_id",
+        compute="_compute_same_siren_siret_partner_id",
         string="Partner with same SIREN",
+        compute_sudo=True,
+    )
+    same_siret_partner_id = fields.Many2one(
+        "res.partner",
+        compute="_compute_same_siren_siret_partner_id",
+        string="Partner with same SIRET",
         compute_sudo=True,
     )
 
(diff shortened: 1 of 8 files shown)


yankinmax (approved):

l10n_fr_siret/models/res_partner.py
```diff
@@ -85,30 +91,50 @@ def _inverse_siret(self):
...
+        self.ensure_one()
+        domain = [
+            (field, "=", getattr(self, field)),
+            ("parent_id", "=", False),
+        ]
+        if self.company_id:
+            domain += [
+                "|",
+                ("company_id", "=", False),
+                ("company_id", "=", self.company_id.id),
```
yankinmax, line 127: `[("company_id", "in", [False, self.company_id.id])]`

alexis-via:
I don't understand what you are trying to change here.
Current implementation : warn if 2 partners have the same SIREN.
Your implementation : 
1)    if another partner has the same SIRET,
2)    or if another partner has the same SIREN but different SIRET.

As SIRET is just an extension of SIREN, your implementation is the same as the current implementation. So what's the point ?

epanisset:
> I don't understand what you are trying to change here. Current implementation : warn if 2 partners have the same SIREN. Your implementation :
> 
> 1. if another partner has the same SIRET,
> 2. or if another partner has the same SIREN but different SIRET.
> 
> As SIRET is just an extension of SIREN, your implementation is the same as the current implementation. So what's the point ?

Hello Alexis, 
Basically for one of our customer, we want to have a warning if **SIRET** are the same. 
So we have added it to this oca module, but keeping also the existing siren warning as complement. 
A company can have 1 SIREN but multiple different SIRET (one per location)

Camptocamp for example has the same SIREN for Camptocamp Chambery & Paris but 2 different SIRET (one per location). 
We are just going further in the warning details. 

image
image

alexis-via:
I'm really against this change ; it adds complexity for no real benefit.

PR #562, opened 2024-08-20
[14.0][FIX] VAT return: fix Journal entry generated in negative VAT scenario + many improvements

alexis-via:
Add test to check the journal entry in negative scenario and improve tests on the generated journal entry in all scenarios

Full re-implementation of the product/service distribution for autoliquidation VAT accounts: we now support autoliquidation VAT accounts in all journals (error if found in sale journal) and not just purchase journal. When auto computation of product/service distribution is not possible, we trigger a wizard to ask the user.

Add message in chatter if ignoring draft moves.


LaetitiaElabore (approved):
tested and approved

PR #563, opened 2024-08-20
[16.0][FIX] l10n_fr_account_vat_return: fix journal entry generated in negative VAT scenario + many improvements

alexis-via:
Add test to check the journal entry in negative scenario and improve tests on the generated journal entry in all scenarios

Full re-implementation of the product/service distribution for autoliquidation VAT accounts: we now support autoliquidation VAT accounts in all journals (error if found in sale journal) and not just purchase journal. When auto computation of product/service distribution is not possible, we trigger a wizard to ask the user.

Add message in chatter if ignoring draft moves.

This is a forward port of https://github.com/OCA/l10n-france/pull/562/files


LaetitiaElabore (approved):
tested and approved

florian-dacosta (approved):
LGTM tested and use in production

PR #568, opened 2024-09-25
ADD: siren/siret fields are now displayed according to a boolean computed field

damdam-s:


l10n_fr_siret/models/res_partner.py
```diff
@@ -137,6 +137,17 @@ def _address_fields(self):
...
+
+    @api.depends("country_id", "is_company", "parent_is_company")
+    def _compute_show_siret_fields(self):
+        countries_show_siret = self.env["res.country"].search(
+            [("show_siret_fields", "=", True)]
+        )
+        for record in self:
+            record.show_siret_fields = record.country_id in countries_show_siret and (
+                record.is_company or record.parent_is_company
+            )
```
petrus-v: instead to search all countries to display siret, could we write something like this:

```suggestion
        for record in self:
            record.show_siret_fields = record.country_id.show_siret_fields and (
                record.is_company or record.parent_is_company
            )
```

petrus-v (approved):

PR #575, opened 2024-10-17
[16.0][l10n_fr_department] improve search res_partner by department

clementelabore:
added _name_search in res.country.department, to filter partner by departments

### DIFF
--- a/l10n_fr_department/model/res_country_department.py
+++ b/l10n_fr_department/model/res_country_department.py
@@ -4,6 +4,8 @@
 # @author Alexis de Lattre (alexis.delattre@akretion.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import re
+
 from odoo import api, fields, models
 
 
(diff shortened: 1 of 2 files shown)


legalsylvain:
Hi clement, could you elaborate which kind of search your PR is allowing ? I'm not sure to understand. 

thanks !

clementelabore:
With my PR, in Contacts, you can search contacts by departments. 
Ex : You can search : 
* "03" => Only Contacts in "Allier (03)"
* Allier => Only Contacts in "Allier (03)"
* "Allier (03)" => Only Contacts in "Allier (03)"
* Futhermore advanced filters on department works "department not equal 03" for example

clementelabore:
hi @legalsylvain can you validate my PR ? Thx.

legalsylvain (commented):
I think it's dangerous to not call _name_search in the _name_search function.

_name_search should always call super, to avoid to break inheritance mechanism.

alexis-via:
I agree with sylvain: this PR is too dangerous and it not worth the added complexity.

PR #578, opened 2024-10-29
18.0 mig l10n fr siret

fredericgrall:


l10n_fr_siret/models/res_partner.py
```diff
@@ -0,0 +1,167 @@
...
+    _inherit = "res.partner"
+
+    siren = fields.Char(
+        string="SIREN",
+        size=9,
+        tracking=50,
+        help="The SIREN number is the official identity "
+        "number of the company in France. It composes "
+        "the first 9 digits of the SIRET number.",
+    )
```
bealdav, line 23: Hi @fredericgrall did you see https://github.com/odoo/odoo/blob/18.0/addons/l10n_fr/models/res_partner.py#L10 ?

Is there an incidence ?

fredericgrall: @bealdav,I'll take a look. Thanks for your comment.

fredericgrall: @bealdav, i'm look the siret field is overide for add inverse methode

alexis-via:
I propose an alternative migration PR with l10n_fr_siret that doesn't depend on account and l10n_fr_siret_account that depends on account, cf https://github.com/OCA/l10n-france/pull/584

remi-filament:
Hi @fredericgrall thank you for porting this module. Since #584 was merged, I suppose this one can be closed ?

PR #584, opened 2024-11-26
[18.0][MIG] l10n_fr_siret + add glue module l10n_fr_siret_account

### LINKED ISSUE #38: l10n_fr_siret: unrequired dependency on account
It would be great if we could split a bit l10n_fr_siret to remove the dependency on account. 

We have one customer at least who would like to record a SIRET number on partners but does not wish to pay an Enterprise Contract for accounting which he is not going to use.

alexis-via:
Since v18, l10n_fr doesn't depend on account (only on base) and there is a glue module l10n_fr_account.
So, we should follow this movement in OCA and have l10n_fr_siret and l10n_fr_siret_account
Fixes #38 which was opened 9 years ago !!!
Small improvements: add siren/siret/nic in list view as optional=hide

This migration PR is an alternative to #578


l10n_fr_siret_account/models/res_partner.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    siren = fields.Char(tracking=50)
+    nic = fields.Char(tracking=51)
```
legalsylvain, line 12: I don't get why you redefine this two fields on the glue module. Why not in base module ? 
also I'm not very familiar with this new tracking values. could you elaborate ?

legalsylvain:
> Since v18, l10n_fr doesn't depend on account (only on base) and there is a glue module l10n_fr_account.

Very nice modular move.

>  So, we should follow this movement in OCA and have l10n_fr_siret and l10n_fr_siret_account

+1, indeed.

l10n_fr_siret_account/models/res_partner.py
alexis-via: tracking is a feature of the mail module, not base. l10n_fr_siret_account depend on account which depend on mail.
For example, you will not see tracking=X on res.partner fields defined in the base module... they are in the mail module: https://github.com/odoo/odoo/blob/18.0/addons/mail/models/res_partner.py#L19

l10n_fr_siret/models/res_company.py
```diff
@@ -0,0 +1,20 @@
...
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    # siret field is defined in l10n_fr module on res.partner
+    # with an unstored related field on res.company
+    siret = fields.Char(store=True, readonly=True)
+    siren = fields.Char(
+        string="SIREN", related="partner_id.siren", store=True, readonly=False
+    )
+    nic = fields.Char(
```
risto42, line 18: Presuming the nic here seems to be problematic.

The basic problem is that we've only been able to structure the res_partner companies (unique SIREN) by having the partner marked 'company' having the SIREN, but with a null NIC giving a SIRET as <siret>*****, adding for each establishment a child partner with it's NIC set correctly. 

For the case of the HQ (siège), setting the partner type 'invoice' as well as setting it as the default invoiced partner for the company (for other partners, customer or suppliers, it may not be exactly that way).

What happens here with a blank nic is that it'll probably break the SIRET in a number of reports.
Unfortunately, it is seemingly not possible to non-programatically change the partner_id on the company.

Perhaps what is needed on the res_partner company  is a HQ link to the child partner currently registered as HQ (siège), which link could be followed to make res_company happy... or at least so if NIC is null.

BTW, when the company moves [the HQ], it is straight-forward that the company partner is the only place that these details should be managed (creation of establishments, archival of closed establishments, et cetera).

Last but not least, the SIRET has a opening date, and at some point in time a possible closing date.

alexis-via: My advice is:
- on the parent partner: address of the HQ (siège social) with siren + NIC of the HQ
- one child partner with type = other/delivery/invoice for each secondary establishment with a different NIC value

risto42: This really doesn't work stably.  It may be ok with a single permanent registered address for a company but not if the company moves or has secondary addresses.
The way I mentioned is the only way we've found to ensure integrity concerning the auxiliary accounts, not to mention references to any other address usage.
The model was much cleaner when there was still a res.partner.address, because SIREN==res.partner and SIRET==res.partner.address

alexis-via: Just like you, I prefered the previous datamodel with res.partner and res.partner.address !!! :)

PR #594, opened 2024-12-17
[14.0][FIX] l10n_fr_account_vat_return: filter on VAT taxes for France

alexis-via:
This commit adapts the code to the scenario where you have VAT taxes for other countries (EU B2C over 10k€, Polynésie française, ...). It "backports" the field country_id on account.tax which is introduced in the account module in odoo v15.


alexis-via:
Tests needs to be updated to set the country_id field on taxes

PR #617, opened 2025-02-04
[16][account_banking_fr_lcr][FIX]  CFONB file -: take payment date instead of today's date

florian-dacosta:


florian-dacosta:
@alexis-via we currently take todays's date as requested date here : https://github.com/OCA/l10n-france/blob/16.0/account_banking_fr_lcr/models/account_payment_order.py#L182
Because in `account_payment_order` the `account.payment`'s date is today
https://github.com/OCA/bank-payment/blob/16.0/account_payment_order/models/account_payment_line.py#L198

I guess we should do the same as here : https://github.com/OCA/bank-payment/blob/17.0/account_banking_sepa_credit_transfer/models/account_payment_order.py#L87 and take the payment_line_date

alexis-via:
The problem is in account_payment_order, not in the LCR module. The account.payment 's date should be the payment date, not today.
In my branch for bank-payment v16, which the LCR module depend on, the account.payment date is the payment date:
https://github.com/akretion/banking/blob/16-payment_order-improve_and_cleanup/account_payment_order/models/account_payment_line.py#L229

florian-dacosta:
Ok, I'll close this one.

PR #620, opened 2025-02-06
[18.0][MIG] account_balance_ebp_csv_export

alexis-via:


alexis-via:
@njeudy I added support for EBP XLSX export (I kept the CSV export).
I added a note in the readme to say that we recommend XLSX export for Teledec. I kept CSV export because it might still be useful to import the balance file in other liasse fiscale software.

alexis-via:
Tests are red because:
```
ERROR odoo odoo.osv.expression: Non-stored field account.account.group_id cannot be searched
```
this is a problem in the module account_financial_report.

njeudy (approved):
just tested on my prod database with my teledec account: OK!

PR #621, opened 2025-02-10
[IMP] l10n_fr_department: Add department also on company model

legalsylvain:
Hi all.
could you take a look on this trivial PR ? 
thanks !

### DIFF
--- a/l10n_fr_department/model/res_company.py
+++ b/l10n_fr_department/model/res_company.py
@@ -1,0 +1,14 @@
+# Copyright 2025-Today: GRAP (http://www.grap.coop/)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    country_department_id = fields.Many2one(
+        comodel_name="res.country.department",
+        related="partner_id.country_department_id",
+        store=True,
+    )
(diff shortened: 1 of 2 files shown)


l10n_fr_department/model/res_company.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2014-2022 Akretion France (http://www.akretion.com/)
+# author: Alexis de Lattre <alexis.delattre@akretion.com>
```
alexis-via: You can write your name and copyright here :)

l10n_fr_department/model/res_company.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    country_department_id = fields.Many2one(
+        string="Country Department",
```
alexis-via: As it is a related field, it will automatically take the string from the field it points to, so no need to redefine it.

alexis-via (approved):

l10n_fr_department/model/res_company.py
legalsylvain: hehe ! fixed !

l10n_fr_department/model/res_company.py
legalsylvain: indeed ! removed. thanks !

PR #628, opened 2025-02-20
[18.0][ADD] l10n_fr_account_oca: OCA fork of l10n_fr_account

alexis-via:
Since v14, OCA/l10n-france has a module l10n_fr_oca which is a fork of l10n_fr.
In v18, the chart of accounts was completely reorganized and l10n_fr has been split between l10n_fr and l10n_fr_account. So the best solution was to create a new fork of l10n_fr_account to l10n_fr_account_oca.


l10n_fr_account_oca/data/template/account.account-fr_oca.csv
```diff
@@ -0,0 +1,675 @@
+id,name,code,account_type,tag_ids,reconcile
+pcg_1011,Capital souscrit - non appelé,101100,equity,,False
```
legalsylvain: Hi. could you avoid to change the format of the file regarding the original one ? 
(https://raw.githubusercontent.com/odoo/odoo/refs/heads/18.0/addons/l10n_fr_account/data/template/account.account-fr.csv)
I mean : 
- do not change delimiter
- do not remove column

So it is allways easy to compare / meld both files. (the original ones, and the forked one).

thanks !

legalsylvain (changes requested):
Hi @alexis-via.

thanks for porting this module.

A format remark inline.

Otherwise, LGTM.

Thanks !

florian-dacosta:
Fix pre-commit + differenciate native and OCA CoA in the configuration list

https://github.com/akretion/l10n-france/pull/13

l10n_fr_account_oca/data/template/account.account-fr_oca.csv
alexis-via: For this specific file account.account-fr_oca.csv, my starting point was NOT the file from l10n_fr_account, but I took the file from OCA/l10n-france/l10n_fr_oca/data/account.account.template.csv v16
As you can see in https://raw.githubusercontent.com/OCA/l10n-france/refs/heads/16.0/l10n_fr_oca/data/account.account.template.csv
the original file doesn't have any quoting ; that's why my file doesn't have quoting.

The delimiter is always coma, so I didn't change the delimiter.
I removed the column name@fr on purpose. In v18, the "name" of account.account is translatable. Translating the account names is quite a difficult task ; instead of having a bad automatic translation certainly full of mistakes, I prefer not to have any translation at all.

legalsylvain: > I removed the column name@fr on purpose. In v18, the "name" of account.account is translatable. Translating the account names is quite a difficult task ; instead of having a bad automatic translation certainly full of mistakes, I prefer not to have any translation at all.

Yes, I saw. I think it's legit, and reduce the maintenance time.

> For this specific file account.account-fr_oca.csv, my starting point was NOT the file from l10n_fr_account, but I took the file from OCA/l10n-france/l10n_fr_oca/data/account.account.template.csv v16

hum. I see it a little problematic. Is it possible to fork from l10n_fr_account file ? Otherwise, it is just not possible to easily meld both files, so, I don't understand how to maintain the module correctly over time.

Or did I missed something ?

florian-dacosta:
@alexis-via 
The module does not install correctly anymore with demo data.
It seems to be linked to this recent change : https://github.com/odoo/odoo/commit/925f8cbe6aebad20de5423814b43e4c07644e9e1

Here is the error message I got : 
```
odoo.tools.convert.ParseError: while parsing /odoo/links/l10n_fr_account_oca/demo/demo_company.xml:38
Tax Payable Account (445510 TVA à décaisser) should be an account of type "Receivable" or "Payable" with both options "Allow Reconciliation" and "Non Trade" enabled.
```

alexis-via:
I propose something: remove accounts 512100 and 531100 from l10n_fr_account_oca/data/template/account.account-fr_oca.csv
because equivalent accounts are created automatically when adding a bank journal and a cash journal, so, in the end, these 2 accounts always stay unused.

hparfr:
>I propose something: remove accounts 512100 and 531100 from l10n_fr_account_oca/data/template/account.account-fr_oca.csv

I'm ok

florian-dacosta (approved):

legalsylvain:
https://github.com/OCA/l10n-france/pull/628#discussion_r1987309083 still requiring answer. otherwise, LGTM.

PR #630, opened 2025-02-25
[14.0][IMP] l10n_fr_siret: improve data migration script

alexis-via:
Improve data migration for fields siret/siren/nic: if we have a valid siren inside a bad siret, we keep the siren.


florian-dacosta (approved):
LGTM (code review)

PR #632, opened 2025-03-03
[18.0][IMP] l10n_fr_siret: M2M for duplicate partners

alexis-via:
The warning banner for duplicate siren partners is now a many2many instead of a many2one, taking advantage of the new widget introduced in v18.


alexis-via:
dupli

PR #633, opened 2025-03-12
[18.0][MIG] l10n_fr_chorus_account: Migration to 18.0

widadelbekkouri-odoo:


alexis-via:
The migration is already here: https://github.com/OCA/l10n-france/pull/602

PR #639, opened 2025-03-25
[16.0][FIX] l10n_fr_intrastat_product: Incompatibility with other intrastat

juancarlosonate-tecnativa:
There is incompatibility if you want to have multiple Intrastat. I believe this change solves it, but i need help reviewing it.

@tecnativa TT55417
@pedrobaeza @victoralmau @alexis-via


alexis-via:
@juancarlosonate-tecnativa Could you explain more in depth the problem you are facing ?
I don't want to solve it that way.
Looking rapidly at the source code, one point can cause problem: the inherit of _prepare_invoice_domain() should only modify the domain for FR intrastat, not all. We need to fix this.

juancarlosonate-tecnativa:
@alexis-via I'm trying to have both `l10n_es_intrastat_report` and `l10n_fr_intrastat_product` at the same time, and there are incompatibilities when, for example, creating an intrastat.product.declaration. I'm not sure how to solve this, do you have any ideas? Thanks!

victoralmau:
> @alexis-via I'm trying to have both `l10n_es_intrastat_report` and `l10n_fr_intrastat_product` at the same time, and there are incompatibilities when, for example, creating an intrastat.product.declaration. I'm not sure how to solve this, do you have any ideas? Thanks!

I think we will have to analyze in detail the existing problem, but there are several possibilities to make both compatible, for example, in the “problematic” methods, check if the company uses a Spanish or French chart of accounts, otherwise, do not execute the piece of code that refers to the specific localization.

alexis-via:
@juancarlosonate-tecnativa please be more specific and tell us the exact details
@victoralmau I fully agree with you ; this is the solution !

PR #645, opened 2025-04-24
[16.0][FIX] l10n_fr_pos_caisse_ap_ip: remove BF tag

alexis-via:
Remove BF tag which cause issues with newer Ingenico software currently being deployed.

Thanks to https://forum.pcsoft.fr/fr-FR/pcsoft.fr.windev/247762-tpe-protocole-concert/read.awp for the solution!


alexis-via:
I received the confirmation this morning from the Grande Chartreuse museum that this PR fixes the problem with Ingenico desk/5000. So let's merge this rapidly.

gregory-moka:
Hello @alexis-via (Hope you're doing well).

This is a suggestion for your next PR on this addon, maybe it could be usefull to show the protocol version handled by this addon on the readme file (here, Concert v118) ?
We had the same case on several museum we have as customers :)

Thanks for the PR

PR #649, opened 2025-04-25
[17.0] Chorus: source for chorus service can be easily inherited + fix when a flow has more than 10 invoices

alexis-via:
Don't hard-code chorus service source from invoice_partner.fr_chorus_service_id. Allows to have another source for
 chorus service by inheriting l10n_fr_chorus_account.


alexis-via:
Fwport of https://github.com/OCA/l10n-france/pull/605 from v16 to v17

PR #653, opened 2025-05-07
[17.0][REM] remove broken symbolic link in setup

vincent-hatakeyama:
The module does not exists and the directory is not needed since the use of whool to package addons.


vincent-hatakeyama:
Some tools do not like broken symbolic links, for example carvel-dev/vendir#418.

rvalyi (approved):

rvalyi:
trivial, merging...

PR #655, opened 2025-05-10
[14.0] l10n_fr_das2: move declaration threshold to pyfrdas2

alexis-via:
Move the partner declaration threashold to the pyfrdas2 lib. This configuration parameter should never be changed by the use anyway. With the encryption that change every year, we will release every year a new lib, so we can change the result of get_partner_declaration_threshold() if the tax administration changes the amount.


alexis-via:
Seems that the new stdnum 2.0 release on may 5 2025 is not compatible with python 3.6.
Unfortunately, tests for 14.0 are running on python 3.6.

rvalyi:
> Seems that the new stdnum 2.0 release on may 5 2025 is not compatible with python 3.6.
> Unfortunately, tests for 14.0 are running on python 3.6.

You could run the tests with Python 3.8 by using the ghcr.io/oca/oca-ci/py3.8-odoo14.0:latest image in .github/workflows/test.yml (we do that in l10n-brazil), but that would likely mean adjusting other tests that may fail on 3.8...

EDIT: @alexis-via so actually it seems trivial to get the CI for 14.0 running on Python 3.8. I did a PR here https://github.com/OCA/l10n-france/pull/656  cc @sebastienbeau

PR #658, opened 2025-05-22
[18.0][ADD] country_fr module: allow to load french CoA without manual settings

bealdav:
```
Apply French Chart of Account to main company when installed before `account` module.

To achieve this target current module set FR country `base.main_company` on partner.

Account module will do the rest of the job.

It has ne effect if `account` module is already installed.
```

cc @florian-dacosta @legalsylvain @alexis-via @sebastienbeau


country_fr/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "name": "Country Fr",
+    "summary": "Set FR country `base.main_company` on partner.",
+    "version": "18.0.1.0.0",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "development_status": "Beta",
+    "category": "Localisation",
+    "website": "https://github.com/OCA/l10n-france",
+    "license": "AGPL-3",
+    "depends": [
+        "base",
```
legalsylvain, line 12: Doing that, you force all users that pull this repo to install this module. 
I dont think it's OK. 
At a first sight, i see 2 options.
- set auto install to False, and auto install it via OCA module module_change_auto_install (https://github.com/OCA/server-tools/tree/16.0/module_change_auto_install)
- put that module in a isolated repo.

Do you see other options ?

legalsylvain (changes requested):

country_fr/__manifest__.py
bealdav: Or a third one, remove auto_install

bealdav: Done

country_fr/README.rst
```diff
@@ -0,0 +1,88 @@
...
+
+Apply French Chart of Account to main company when installed before
+``account`` module.
+
+To achieve this target current module set FR country
+``base.main_company`` on partner.
+
+Account module will do the rest of the job.
+
+The module is auto installed.
```
legalsylvain: Should be removed.

country_fr/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,7 @@
+Apply French Chart of Account to main company when installed before `account` module.
+
+To achieve this target current module set FR country `base.main_company` on partner.
+
+Account module will do the rest of the job.
+
+It has ne effect if `account` module is already installed.
```
legalsylvain:
```suggestion
It has no effect if `account` module is already installed.
```

legalsylvain (commented):
Not a blocking point. I think that a mention to module module_change_auto_install should be written.

Otherwise, lgtm

legalsylvain (approved):

country_fr/partner.xml
```diff
@@ -0,0 +1,5 @@
+<odoo>
```
florian-dacosta: I think it would be better to have a noupdate here

florian-dacosta (approved):
LGTM just a minor comment

country_fr/partner.xml
bealdav: Done

bealdav:
Here is the fail case
`The failing job indicates issues related to the use of the pkg_resources module, which is deprecated. This warning appears in the Python file pyfrdas2.py in the module pyfrdas2. The error log also contains a reference to the Odoo module loading process, suggesting that the failure occurred during module initialization.`

PR #660, opened 2025-06-05
14.0 chorus improve

sebastienbeau:
@alexis-via


alexis-via:
The fix for the "more than 10 invoices bug" is already merged in v14 (another implementation) via this PR https://github.com/OCA/l10n-france/pull/651

alexis-via:
Discussion with @sebastienbeau : the logger.warning for multi-pages that I added in PR #651 could be converted to a raise

PR #667, opened 2025-07-11
[16.0][FIX] l10n_fr_intrastat_product: Remove error blocking region code when warehouse has no partner

juancarlosonate-tecnativa:
@tecnativa TT55417
@victoralmau @alexis-via @pedrobaeza 

In cases where there's no partner_id for example, in a virtual warehouse it should fall back to the one you've established as the priority, which is the company's partner country_department_id. Is that correct, @alexis-via, according to the French localization requirements?

The UserError i removed was blocking this logic https://github.com/OCA/l10n-france/blob/210d2f45e1c3e6bb9087db3ec8d9bb95a0a599e2/l10n_fr_intrastat_product/models/intrastat_product_declaration.py#L97 and forcing the use of partner_id. 

And if the partner’s department cannot be found on the company either, it will then fall back to this logic https://github.com/OCA/l10n-france/blob/210d2f45e1c3e6bb9087db3ec8d9bb95a0a599e2/l10n_fr_intrastat_product/models/intrastat_product_declaration.py#L98 which was also broken.

### DIFF
--- a/l10n_fr_intrastat_product/models/stock.py
+++ b/l10n_fr_intrastat_product/models/stock.py
@@ -2,8 +2,7 @@
 # @author Alexis de Lattre <alexis.delattre@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import _, models
-from odoo.exceptions import UserError
+from odoo import models
 
 
 class StockWarehouse(models.Model):
@@ -11,9 +10,7 @@ class StockWarehouse(models.Model):
 
     def _get_fr_department(self):
         self.ensure_one()
-        if not self.partner_id:
-            raise UserError(_("Missing partner on warehouse '%s'.") % self.display_name)
-        return self.partner_id.country_department_id
+        return self.partner_id.country_department_id if self.partner_id else None
 
 
 class StockLocation(models.Model):


l10n_fr_intrastat_product/models/stock.py
```diff
@@ -2,18 +2,15 @@
...
+from odoo import models
 
 
 class StockWarehouse(models.Model):
     _inherit = "stock.warehouse"
 
     def _get_fr_department(self):
         self.ensure_one()
-        if not self.partner_id:
-            raise UserError(_("Missing partner on warehouse '%s'.") % self.display_name)
```
pedrobaeza, line 15: I think this error should be raised at other level, once you don't have the region via partner nor specific field.

juancarlosonate-tecnativa: ok, but i guess this check should be at the intrastat_product level, since it's ultimately the one not verifying whether region_code can be false, right?

pedrobaeza: It depends if this region is mandatory only for France or for any declaration.

juancarlosonate-tecnativa: Well, in that case, it's just as i mentioned in the PR: if country_department_id is the fallback after not finding it on the partner as per the French declaration logic then, if it's also missing on the partner company, it will show the message: https://github.com/OCA/l10n-france/blob/210d2f45e1c3e6bb9087db3ec8d9bb95a0a599e2/l10n_fr_intrastat_product/models/intrastat_product_declaration.py#L97

pedrobaeza (approved):

l10n_fr_intrastat_product/models/stock.py
alexis-via: I'm ok to remove the raise here. We could replace the raise by writing the same message in the chatter.

pedrobaeza: Afaik, there's already a raise following the flow, so the message post will also be rollbacked.

juancarlosonate-tecnativa: But if we put it in the chatter, then whenever there’s a warehouse where, for some reason, we don’t want to set the partner, the message will still appear in the chatter and that might be confusing for the user, right?

alexis-via (approved):

PR #674, opened 2025-08-02
[12.0][FIX] account_banking_fr_lcr: do not search on non stored acc_type field

remi-filament:
When trying to generate payment_lines, you get the following error :
`Non-stored field res.partner.bank.acc_type cannot be searched.`


alexis-via (approved):

remi-filament:
It seems merge was interrupted, can anyone from @OCA/local-france-maintainers start it again ?
Thanks !

PR #677, opened 2025-09-02
[IMP] l10n_fr_fec_oca: work with l10n_multilang installed

len-foss:
With l10n_multilang, fields are jsonb columns. This is complicated at it supports both case, but in 18.0 it seems fields are translated in the base module and so this can be simplified there.


len-foss:
@alexis-via I've also had the request to mangle the account name with the partner id for some accountant. 
Given the way the code is written, it can't be added by inheritance (unless we add very ad-hoc hooks). Is it something that is common enough to be added to this module? Otherwise I'll just work on a fork for that customer.

PR #679, opened 2025-09-03
FIX chorus_account: remove call to _get_new_token because has result have no clear_cache attribute

bealdav:
FIX chorus_account: remove call to _get_new_token because has result have no clear_cache attribute


florian-dacosta:
Fixed by https://github.com/OCA/l10n-france/pull/750

PR #542, opened 2024-04-25
[17.0][MIG] l10n_fr_siret_lookup

David-Luis-Mora:

Issue #591, opened 2024-12-11
problem occuring during VIES validation using search via API SIRENE


risto42:
Unfortunately being a victim of success, VIES is frequently overloaded returning the error message MS_MAX_CONCURRENT_REQ which fails the query, failing the operation.
```
2024-12-10 13:07:06,569 1132 WARNING myODOO zeep.wsdl.bindings.soap: Forcing soap:address location to HTTPS 
2024-12-10 13:07:11,579 1132 ERROR myODOO odoo.addons.l10n_fr_siret_lookup.models.res_partner: VIES query failed: HTTPSConnectionPool(host='ec.europa.eu', port=443): Read timed out. (read timeout=5) 
2024-12-10 13:07:11,605 1132 WARNING myODOO odoo.http: Erreur pendant l'interrogation du VIES.
Erreur technique : HTTPSConnectionPool(host='ec.europa.eu', port=443): Read timed out. (read timeout=5). 
2024-12-10 13:07:26,917 1131 WARNING myODOO zeep.wsdl.bindings.soap: Forcing soap:address location to HTTPS 
2024-12-10 13:07:26,965 1131 ERROR myODOO odoo.addons.l10n_fr_siret_lookup.models.res_partner: VIES query failed: MS_MAX_CONCURRENT_REQ 
2024-12-10 13:07:26,984 1131 WARNING myODOO odoo.http: Erreur pendant l'interrogation du VIES.
Erreur technique : MS_MAX_CONCURRENT_REQ. 
```
It would be nice to be able to  'tag' the fact that a VIES check is demanded if the interactive query fails, and eventually schedule a CRON job to retry after some interval of time.

[https://stackoverflow.com/questions/74821299/ms-max-concurrent-req-vies-vat-validation](url)

Then, at least, if one is creating a record, the effort isn't totally lost.

risto42:
recently it is for the most part impossible to use this function, please upgrade from _enhancement_ to **bug**
it should therefore be optional (by default) to do the VIES validation.  

In any event, if base_vat is installed, `env.company.vat_check_vies` should be respected, or at least this value should be used as the default value of an eventual option specifying to do a synchronous VIES check.
```
There are two different levels of VAT number validation:
--------------------------------------------------------
    * By default, a simple off-line check is performed using the known validation
      rules for the country, usually a simple check digit. This is quick and 
      always available, but allows numbers that are perhaps not truly allocated,
      or not valid anymore.

[...]

PR #680, opened 2025-09-22
[17.0][MIG] l10n_fr_siret_lookup : Migration to 17.0

remi-filament:
Take over #542 which was unattended and closed by lack of activity.


remi-filament:
Hi @alexis-via would you mind having a look at this one ?
I am also wondering if we should not change a bit the behaviour about VIES validation, since it is very often ending with "VIES query failed: MS_MAX_CONCURRENT_REQ" as reported in #591 

When using the wizard, if VIES check fails, you cannot import the line, you get a UserError and you are blocked there. We should allow of importing at least all other fields but vat.
On VIES website, they propose to move to async request : https://viesapi.eu/vies-error-the-maximum-number-of-concurrent-requests-has-been-reached-fr/

I am not sure it makes sense here, but maybe we could add 2 configuration parameters allowing : 
1. Enable / Disable check of VIES
2. In case VIES check is disabled or timesout : take vat computed from SIREN (you will probably end up with a few incorrect VAT numbers) or do not compute vat (as done today)

For the timeout case we could probably raise a warning or notifification that this was not checked (or add a warning icon next to VAT ?) and maybe a button to recompute / check VAT number against VIES OR a cron job that would recompute all the ones marked with the warning icon (to get an async behaviour as suggested by VIES website) ?

What do you think ?

julienCozo (approved):

Hugo-Trentesaux (approved):

flotho (changes requested):
Hi, sadly it lloks like that the url is no more available
please check https://community.huwise.com/gather-data-67/a-new-marketplace-to-better-showcase-useful-data-629?fid=67&tid=629
and https://hub.huwise.com/explore/?orderBy=updated_at+DESC

petrus-v:
Just trying the module on a test server and I'm getting the following error while selecting the company from the wizard, I wondering if it's related to the previous message regarding changes in APIs

image

It's working putting the siret directly in the name field and changing the focus to the next field.

remi-filament:
Bonjour @petrus-v ,

I just tested and opendatasoft API is still working,
Your error seems related to library stdnum calling check_vies() which calls get_soap_client().
What stdnum version are you using ?
Odoo asks for 1.17 or 1.19 in requirements.txt of vestion 17 (https://github.com/odoo/odoo/blob/3d260f405d72f463722bd5e5be279c59baf77334/requirements.txt#L68-L69)

This verify parameter seems to have been added in v2.0
However there is a default value for this parameter so I am not sure why it would make you query fail...

remi-filament:
Julien has shared in https://github.com/OCA/l10n-france/pull/741 the implementation of the same functionalities using INPI API (which requires an account when SIRENE did not).

PR #681, opened 2025-09-22
[18.0][MIG] l10n_fr_siret_lookup: Migration to version 18.0

remi-filament:
Replaces #579 which was unattended and closed because of lack of activity.
Started from #680

PR #682, opened 2025-09-23
[18.0][IMP] l10n_fr_siret_lookup: Conditionally check VIES and force VAT + tests

remi-filament:
As proposed in https://github.com/OCA/l10n-france/pull/680#issuecomment-3317737379

I have added a couple of things on this module :
1. Add dependency to base_vat (needed for 2. below + base_vat is a dependency of l10n_fr_account and most of localizations so it should be installed anyway if using Invoicing and French localization)
2. Check VIES server for computed VAT number only if company is configured to check VAT number against VIES (field vat_check_vies from base_vat module)
3. Add an extra config parameter (company dependant) : force_vat_siret_lookup, allowing to use VAT number computed by this module from SIREN even if VIES check fails (for timeout) or if VIES check is disabled
4. Add tests for testing the different possible scenarios for all onchange

This is also linked to #591 trying to workaround the MS_MAX_CONCURRENT_REQ raised by VIES service most of the times when checking for FR VAT.

Note that if VIES returns that VAT is invalid, the VAT is emptied in any scenario (independent of value of force_vat_siret_lookup)

We could go a few steps further, for instance by showing the validation status on partner form view and either allowing to rerun the check or to be handled by cron.


remi-filament:
Built on top of https://github.com/OCA/l10n-france/pull/681

remi-filament:
Hi @alexis-via I would appreciate your feedback on this proposal when you can spare a few minutes !
Thanks

julienCozo (approved):

remi-filament:
Thinking again about this module how to request new validation from VIES service, I started another module not directly related to this one, since the issue with max request happens also without this module installed.
See https://github.com/OCA/account-financial-tools/pull/2154

Hugo-Trentesaux (approved):

flotho:
Hi @remi-filament , any chance to rebase to trigger CI/CD and go to merge

remi-filament:
Done @flotho although I am still looking for a replacement regarding your comment here : https://github.com/OCA/l10n-france/pull/680#pullrequestreview-3365967151
We have implemented something similar with INPI RNE API or INSEE SIRENE API but I need to go further into the analysis...

alexis-via (approved):

alexis-via:
Yes, it's a pain that VIES very very often fails on FR VAT numbers ; I have experienced that too.
I don't have any better idea to work around this problem. Let's go with that, and we'll see with the real-life experience if it's ok for not...

PR #692, opened 2025-11-06
[18.0][IMP] l10n_fr_account_payment_intl_credit_transfer: add default regulatory reporing on partners

alexis-via:
reg_rep

Raise an error if a non-SEPA credit transfer is confirmed with a payment above 50k€ without regulatory reporting.


alexis-via:
This PR new depend on https://github.com/OCA/bank-payment-alternative/pull/46

@rlemeur The 2 bugs we reproduced together are fixed in the last commit.

rlemeur (approved):

PR #698, opened 2025-11-11
[ADD] l10n_pf : add Fiscal Localization module for French Polynesia

metaminux:
Hi all,

Not sure if this module should be hosted here...

It is strongly inspired by `l10n_fr_oca` though.
The first goal is to provide chart of accounts and taxes to be able to fill DECL 1010 for VAT in French Polynesia.

I'm defining an `account.tax.report` here to be used in a further module `l10n_pf_mis_reports` (I'll use OCA/mis-builder-contrib#19 from @sbidoul to generate the report template)

Any advice, remark, review would be appreciated...

@alexis-via @cvinh 

During development I also discovered a repository from @jburckel... wanna help ?


cvinh:
@metaminux maybe you contact me...

metaminux:
> @metaminux maybe you contact me...

Sure, I will :)

PR #717, opened 2025-12-20
[16][FIX] l10n_fr_fec: more robust/exhautive way to get rid of invisible caracters (like line break)

florian-dacosta:
Recently I had a problem with invalid FEC (not read correctly by excel not valid with a fec validation tools, because of the presence of vertical tab (\x0B)
This PR aims to get rid of this kind of characters in a more exhaustive manner


florian-dacosta:
@alexis-via

Issue #711, opened 2025-12-10
[19.0] l10n_fr_siret: 'siret' field removed in l10n_fr / switch to company_registry


alexis-via:
While working on the migration of l10n_fr_siret to 19.0, I found out that a major change has occurred in the l10n_fr module in 19.0, cf this commit of december 2024 https://github.com/odoo/odoo/commit/6730282b689939eca9df0c811ca3fd015cc48eae
The editor has removed the "siret" field on res.company and res.partner ; he recommends to use the field company_registry instead. In the form view of partners, the company_registry field appears under the label "Siret":

<img width="958" height="736" alt="Image" src="https://github.com/user-attachments/assets/621f8d98-6e67-4afe-a5dd-4c71c85e5645" />

The field "company_registry" appears when your select the country on res.partner (it is hidden if the country is not set).

This is a major change ! The l10n_fr module has the 'siret' field on res.company since Odoo v7 and on res.partner since v11.

We have to decide if we adopt the new datamodel in l10n_fr_siret i.e. use the field "company_registry" to store the SIRET or if we continue with our historic datamodel with the field 'siret' (and 'siren' and 'nic' on res.partner).

Advantage of adopting the native datamodel:
- be compatible with the other apps that use the native datamodel
- using the field company_registry on res.partner to store the SIRET (field defined in the "base" module) would probably avoid several glue modules such as l10n_fr_business_document_import, l10n_fr_account_invoice_import_facturx, l10n_fr_account_invoice_import_simple_pdf, l10n_fr_account_invoice_facturx, etc...

My concerns/questions about this new datamodel:
- how to handle foreign companies that have a SIRET number (cf https://www.expand-cpa.com/siret-pour-entreprise-etrangere-les-etapes-pour-lobtenir/) ; it is not very common, but this scenario can happen.
- what about the fields 'nic' and 'siren' that we have ? I think that the siren field is not very useful ; it could be replaced by 2 methods _get_siren() and _get_siret() and we could imagine that company_registry could accept a valid SIRET with 14 digits or a valid SIREN with 9 digits. In the current implementation of l10n_fr_siret, the "nic" field allowed to support multi-establishment companies with the ability the have the specific "nic" on each address. I don't think it is used in the code itself and I don't think it is widely used... but it was an interesting feature in some advanced scenarios. Maybe we could imagine to have an additionnal module that adds the "nic" field for those that need this advanced modelling.

[...]

PR #723, opened 2025-12-23
[19.0][MIG] l10n_fr_siret

alexis-via:
Implementation proposal following https://github.com/OCA/l10n-france/issues/711 that adopt the new native datamodel : SIRET is now stored on the field "company_registry" of res.partner.


l10n_fr_siret/models/res_partner.py
```diff
@@ -0,0 +1,301 @@
...
+                                "SIREN '%(child_siren)s' of child partner "
+                                "'%(child_partner)s' is different from SIREN "
+                                "'%(parent_siren)s' of its parent partner "
+                                "'%(parent_partner)s'.",
+                                child_siren=siren,
+                                child_partner=rec.display_name,
+                                parent_siren=parent_siren,
+                                parent_partner=rec.parent_id.display_name,
+                            )
+                        )
```
legalsylvain, line 162: Are you sure about this test ? 
I'm asking if some people design "holding" with parent in odoo. if yes, company_registry could be totally different between childs... not sure thought.

legalsylvain (commented):
Thanks for the ping Alexis. I'm asking if we can define siret / siren and nic fields, computed readonly, and stored, with exactly the same function as you wrote.

**Pro :** 
- possibility to make SQL, request, etc.
- lesss useless computation. (those fields are very stable, so no need to recompute at each use).
- possibility to display it in views.

**Cons :** 

Don't see, but maybe !

l10n_fr_siret/models/res_partner.py
alexis-via: In my interpretation of how partners should be organised in Odoo for companies, I consider that 1 parent partner = 1 legal entity (1 personne morale) i.e. a SIREN for a french company (or a VAT number). I think it's a mistake to use the parent / child relationship to consider a holding company which owns other companies created as children companies ; because, in the accounting, only parent companies are considered on journal items, so you end up in accounting to consider that the debt is only on the holding company although some invoices were for some children companies.
But I'm not sure what is the exact vision of Odoo SA on parent/child relations for companies.
This is particulaly important for the e-invoicing reform : when we used to send invoices by email or by post, the most important parameter was the destination email (or postal) address to which the invoice was sent ; but, with the e-invoicing reform, the most important parameter in SIREN or VAT number.

tansadio:
Small fix on top of this migration: missing `mail` dependency breaks `post_init_hook` → akretion/l10n-france#17

vvrossem:
@alexis-via could you include the fix mentioned above in this PR?

yankinmax:
Hello @alexis-via   if you don't mind, I've opened a new PR (to speed up migration process and address the suggestions) (your commits are preserved):

- [ ] https://github.com/OCA/l10n-france/pull/770


@OCA/local-france-maintainers  can you pls close this one?

PR #740, opened 2026-02-10
[18.0][MIG] l10n_fr_account_vat_return + l10n_fr_account_vat_return_teledec

alexis-via:
This is a new migration from v16 to v18: there has been so many changes in v16 recently with the merge of https://github.com/OCA/l10n-france/pull/652 that it was simpler to migrate again.


l10n_fr_account_vat_return/models/l10n_fr_account_vat_return.py
```diff
@@ -0,0 +1,3381 @@
...
+            raise UserError(_("The Journal Entry of this VAT return has been deleted."))
+        if move.state == "cancel":
+            raise UserError(
+                _(
+                    "The Journal Entry of this VAT return has been cancelled. "
+                    "You should set it back to draft."
+                )
+            )
+        if move.state == "draft":
+            move.action_post()
```
sebalix, line 899: In case there is no VAT to report, the move cannot be posted (no lines, returning an error):
```suggestion
            if move.line_ids:
                move.action_post()
            else:
                move.button_cancel()
```
It's blocking as we cannot generate the VAT report of the next month while current one isn't validated.

PR #748, opened 2026-03-10
[16.0][FIX] account_balance_ebp_csv_export: Updated report data method name for balance trial report

Honeyxilia:
In this following account-financial-reporting PR ( https://github.com/OCA/account-financial-reporting/pull/1290 ), the abstract wizard model account_financial_report_abstract_wizard got a new generic method "_prepare_report_data", and as such, the "_prepare_report_trial_balance" was changed to a overloading of the "_prepare_report_data" method, with the same data provided.

As such, the previous "_prepare_report_trial_balance" was no longer available, and using this export action lead to a crash due to a non-found method. So, we update the previous data method to the new one.

PR #749, opened 2026-03-11
[18.0][FIX] account_balance_ebp_csv_export: adapt to changes in account_financial_report 

alexis-via:
This is a sort of forward-port of the fix in 16.0 https://github.com/OCA/l10n-france/pull/748
The code is not exactly the same, but the idea is the same


alexis-via:
The crash was caused by a change introduced in this PR https://github.com/OCA/account-financial-reporting/pull/1408

PR #754, opened 2026-03-20
[19.0][IMP] l10n_fr_intrastat_service: use ElementMaker to generate XML

alexis-via:
I'm evaluating different XML generation method, to find the most readable/pythonic. ElementMaker seems a good option.


alexis-via:
FYI, I'm testing here in this module the new method for XML generation that I plan to use in the e-invoicing reform for France (including the new implementation of Factur-X).

PR #756, opened 2026-03-26
[18.0][FIX] l10n_fr_intrastat_product: adapt to changes in intrastat_product (several fields changed from int to float on decl lines)

alexis-via:
Adapt to changes in intrastat_product (several fields changed from int to float on decl lines) made in this PR:

https://github.com/OCA/intrastat-extrastat/pull/335

As the PR on intrastat_product is already merged, we need to merge it rapidly. But, following my comments after the merge of the PR, a second iteration will probably happen, which may also impact l10n_fr_intrastat_product ; so I propose to merge it after the second iteration.

BTW, the change has also been made in 17.0

### DIFF
--- a/l10n_fr_intrastat_product/models/intrastat_product_declaration.py
+++ b/l10n_fr_intrastat_product/models/intrastat_product_declaration.py
@@ -23,20 +23,18 @@ class IntrastatProductDeclaration(models.Model):
     ]
     _description = "EMEBI"
 
-    # WARNING: will probably be switched to Float in the very near future
-    total_amount = fields.Integer(compute="_compute_fr_numbers")
-
-    @api.depends("declaration_line_ids.amount_company_currency")
-    def _compute_fr_numbers(self):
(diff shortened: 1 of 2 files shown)


l10n_fr_intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -22,22 +23,19 @@ class IntrastatProductDeclaration(models.Model):
     ]
     _description = "EMEBI"
 
+    # WARNING: will probably be switched to Float in the very near future
```
pedrobaeza: As this is overriding the general model, you should remove it, or having this installed will affect all the declarations. We have a customer with Spain and France and both modules installed for example.

alexis-via: done

bguillot (approved):

clementmbr (approved):
LGTM! (code review)

PR #757, opened 2026-03-30
[18][MIG] l10n_fr_hr_rup

florian-dacosta:
Standard migration


l10n_fr_hr_rup/README.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Usage
+=====
+
+Select contracts in list view and print the report to get a PDF.
+
+Known issues / Roadmap
+======================
+
+- Add classical header and legal information in the footer
```
legalsylvain, line 48: I think this one could be done in a few time.

l10n_fr_hr_rup/models/hr_contract.py
```diff
@@ -0,0 +1,38 @@
...
+    employer_address_id = fields.Many2one("res.partner", "Addresse de l'employeur")
+    pcs_id = fields.Many2one(
+        "hr.employee.pcs", "PCS", compute="_compute_pcs_id", store=True, readonly=False
+    )
+    qualification = fields.Char(
+        compute="_compute_qualification", store=True, readonly=False
+    )
+    work_location_id = fields.Many2one(
+        "hr.work.location",
+        string="Localisation du bureau",
```
legalsylvain, line 19: should be english, here.

legalsylvain (changes requested):
Thanks for porting this module.

> This module imports the French official RUP

the description says the module import CSP, but I don't see CSP import.

could you elaborate ? 

thanks !

l10n_fr_hr_rup/README.rst
florian-dacosta: Well, I don't think there is a real need from an existing user/company, I'd rather not do it if not needed, I think it can stay here or remove this section as it seems it is not really in the roadmap anymore.

l10n_fr_hr_rup/models/hr_contract.py
florian-dacosta: I thought it was ok in l10n-fr module. It is actually the case in multiple place in this module.
If it is definitly not tolerated in such a module, I'll change all the french  string

legalsylvain: > I thought it was ok in l10n-fr module.

I don't remember such rule. do you have other exemple ? 

For me, it's just in XSD template that it is allowed.

florian-dacosta:
> Thanks for porting this module.
> 
> > This module imports the French official RUP
> 
> the description says the module import CSP, but I don't see CSP import.
> 
> could you elaborate ?
> 
> thanks !


The description may be a bit misleading, it says "imports the RUP (Registre Unique du Personnel)", which actually means the module allow to print the RUP (report), which contain a column with the CSP. 
But there is not CSP import, you have to add the one you need for the report manually.

legalsylvain:
OK. could you reword ? removing the "import" word, that in Odoo, means that data are imported in database. thanks !

l10n_fr_hr_rup/models/hr_contract.py
florian-dacosta: I don't really have modules in mind nor do I think there is a written rule. After a quick look I found another example : https://github.com/OCA/l10n-france/blob/18.0/l10n_fr_das2/models/res_partner.py
I did not check modules in other l10 repo.
But anyway, If you think it is really better, I'll change + translate, no problem.

legalsylvain: Hi. Well, I don't want contributor to waste time. I know that OCA rules, says all should be written in english. However It could make sense to have an exception.
But, in that case, (and in your l10n_fr_das2, it is not in french. some text are in french, other are in english. As a result, we still have to use weblate.
and weblate, will ask translation for french sentences.

so, I think I prefer all in english that "mixed languages"
(and I'm open to propose an exception in OCA rules. if you think it's interesting. Feel free to open a PR, and ping me).

florian-dacosta: I don't really think  it is worth making an exception on my side. I did keep the french string at first as this is a simple/standard migration.
I'll change the french string so we can move forward

florian-dacosta: @legalsylvain 
Problem is that some stuff don't really make sense to translate (in english I mean), Like REGISTRE UNIQUE DU PERSONNEL, which is the official name of the document, etc...
So we'll always have kind of mixed languages (or worse, fully english with bad /not existing translation

Here, All the report is in french, I feel some translation (in english) won't be very good.
So the question is, is it better to have the module hardcoded in french, or have it in bad english with french translation ?

legalsylvain: could you take a look in OCA issues ? I'm sure the debate has already be done ! 
thanks !

florian-dacosta: I tried to have a look, without success (l10n-france, mainteners-tools, google)...
I guess we can wait a bit to see if there is more opinion about this.

I will translate if needed, but it seems kind of useless, mainly for the RUP report, which as no sense in english.

PR #763, opened 2026-04-14
[18.0] EMEBI 2026 requires a fake VAT for B2C regime 29

alexis-via:
That's for you @florian-dacosta


florian-dacosta (approved):
Thank you
It works as expected.

PR #766, opened 2026-04-17
[IMP] l10n_fr_siret: configurable duplicate warning (SIREN/SIRET)

len-foss:
It is reasonable to have different top-level partners with the same SIREN, while same SIRET would be a duplicate.
This allows to choose what duplicate check should be performed.

### DIFF
--- a/l10n_fr_siret/__init__.py
+++ b/l10n_fr_siret/__init__.py
@@ -1,2 +1,3 @@
 from . import models
+from . import wizard
 from .post_install import set_siren_nic
--- a/l10n_fr_siret/__manifest__.py
+++ b/l10n_fr_siret/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     "name": "French company identity numbers SIRET/SIREN/NIC",
-    "version": "16.0.1.2.0",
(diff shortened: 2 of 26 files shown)


alexis-via:
Users that have several different parent partners with same SIREN will be confronted to this SQL constraint:
https://github.com/odoo/odoo/blob/16.0/odoo/addons/base/models/res_bank.py#L89

I think it is a bad practise and I wouldn't recommend that.

I must say I'm not in favor of this PR. If you want it, you can still change the behavior in another module.

len-foss:
Hello @alexis-via 

Thank you very much for your comment.
The use case that motivated it treats the different entities as separate, because this is how they operate with regards to every business flow (sales, contact partners, payment, etc). So they would be top-level partners.

With regards to the constraint you mention, it would not cause any problem since the constraint is per `partner_id`. But I guess their use differs a bit too much from the standard flow you want to support.

If that's fine with you, I would let the PR open for a while; if somebody else has the same need, then I would separate in another module if needed (and I guess refactor a tiny bit  to make the override easier?). If not then it will close and I'll just user a forked version for the customer who needs it.

PR #768, opened 2026-04-23
[18.0][IMP] l10n_fr_siret: add get methods similar to v19

alexis-via:


remi-filament (approved):
Thanks @alexis-via 
Might be worth explaining what these functions will be used for, so that we avoid that someone thinks they are not called anywhere and proposes to remove them !

sebalix (approved):

flotho (commented):
> Thanks @alexis-via Might be worth explaining what these functions will be used for, so that we avoid that someone thinks they are not called anywhere and proposes to remove them !

Hi @remi-filament , that will be exactly my question, where are those methods used ? I didn't find them in the base code.
@alexis-via , a little will be useful.

Code Review LGTM,

Simple comment

alexis-via:
They are used here for example: https://github.com/akretion/fr-einvoicing/blob/18.0/l10n_fr_einvoicing/models/res_partner.py#L147

flotho (approved):
thanks for all the clarifications

PR #777, opened 2026-05-29
[18.0][IMP] l10n_fr_pos_caisse_ap_ip: add support for ANCV (chèques vacances)

alexis-via:


alexis-via:
Next step : test this new feature on a real Ingenico card reader with support for Chèques vacances.

PR #687, opened 2025-10-31
[18.0][MIG] l10n_fr_fec_oca: Migration to 18.0

P-H-Phuc:
Export report in the following format:
image

PR #783, opened 2026-06-11
[19.0][MIG] l10n_fr_fec_oca: Migration to 19.0

pierrelauret:
Migration of `l10n_fr_fec_oca` from 18.0 to 19.0.

Based on the 18.0 work from #687 (@P-H-Phuc).

Changes for the 19.0 migration:
- Bump version to 19.0.1.0.0
- The parent wizard `l10n_fr.fec.export.wizard` (from `l10n_fr_account`)
  removed the `exclude_zero` field in 19.0 and now always excludes
  zero-balance lines in its base domain. Adapted `_get_base_domain()`
  accordingly to stay consistent.
- Replaced deprecated `self._cr` by `self.env.cr`
- Replaced percent-formatted URL by an f-string (ruff UP031)

Tested: module installs and FEC generation works on a 19.0 database
with posted entries.

@P-H-Phuc @alexis-via FYI


pierrelauret:
The `l10n_fr_fec_oca` tests now pass.

The remaining CI failure comes from `l10n_fr_intrastat_service`
(`ValueError: External ID not found: product.product_product_1` in its
`setUpClass`), which is out of scope for this PR — that module is not
yet migrated to 19.0 and its test relies on demo data removed in 19.0.
This PR only migrates `l10n_fr_fec_oca`.

PR #784, opened 2026-06-17
[14.0][FIX] l10n_fr_das2: drop obsolete config settings view on upgrade

klodr:
## Problem

`l10n_fr_das2` v14.0.4.0.0 removed the `fr_das2_partner_declare_threshold` setting field (the declaration threshold now comes from the `pyfrdas2` lib) together with its `res.config.settings` view `view_account_config_settings` — but **without a migration** to clean up the stale view record left in the database.

On upgrade from an earlier version, as soon as any other module re-renders the `res.config.settings` form (e.g. a module extending the accounting/general settings), view validation fails with:

> ValidationError: Field 'fr_das2_partner_declare_threshold' does not exist in model 'res.config.settings'

and the registry fails to load (the whole `-u` rolls back). Odoo would normally drop the orphan view in `_process_end`, but the crash happens earlier, while reloading the views.

## Solution

Add a `pre-migration.py` (version bumped to `14.0.4.0.1`) that drops the obsolete view **before** the views are reloaded:

- `l10n_fr_das2/migrations/14.0.4.0.1/pre-migration.py`

## Testing

Reproduced and fixed on a real production database (Odoo 14.0, `-u all`): without the cleanup the upgrade crashes on the obsolete view; removing it beforehand lets the upgrade complete with no error.

### DIFF
--- a/l10n_fr_das2/__manifest__.py
+++ b/l10n_fr_das2/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "DAS2",
-    "version": "14.0.4.0.0",
+    "version": "14.0.4.0.1",
     "category": "Invoicing Management",
     "license": "AGPL-3",
     "summary": "DAS2 (France)",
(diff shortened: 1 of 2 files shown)


klodr:
Note for maintainers: the failing `pre-commit` check here is unrelated to this PR's content. The `setuptools-odoo-make-default` and `setuptools-odoo-get-requirements` hooks crash at import with:

`ModuleNotFoundError: No module named 'pkg_resources'`

This is an environment regression (recent `setuptools` ≥ 81 dropped `pkg_resources`, which `setuptools_odoo` still imports), affecting any PR run at the moment — not this change. All other hooks pass (black, flake8, pylint-odoo, isort, prettier, …), and `test with OCB` / `test with Odoo` are green.

PR #785, opened 2026-06-17
[19.0][MIG] l10n_fr_mis_reports: Migration to 19.0

aurelien-tempo:
No changes to the data files as there were no changes to the models in the mis_builder module.
I cleaned up the git history a bit, as suggested in the OCA guidelines.

Pinging @alexis-via as you seem to be the maintainer of this module and did the previous migrations.


pilarvargas-tecnativa (approved):

pilarvargas-tecnativa:
@Tecnativa TT61945 

@eduezerouali-tecnativa please review

eduezerouali-tecnativa (approved):

aurelien-tempo:
Thank you for the review

l10n_fr_mis_reports/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    "name": "MIS reports for France",
+    "version": "19.0.1.0.0",
+    "category": "Accounting & Finance",
+    "license": "AGPL-3",
+    "summary": "MIS Report templates for the French P&L and Balance Sheets",
+    "author": "Akretion,Odoo Community Association (OCA)",
+    "maintainers": ["alexis-via"],
+    "website": "https://github.com/OCA/l10n-france",
+    "depends": ["mis_builder"],
```
victoralmau, line 14: It's strange that it doesn't have the `l10n_fr` dependency.

victoralmau (approved):

l10n_fr_mis_reports/__manifest__.py
pedrobaeza: @aurelien-tempo can you check this?

aurelien-tempo: Will check this week when I find the time

aurelien-tempo: @victoralmau I checked, the module does not use anything from l10n_fr. So a dependency would be a "logical" one and not a "technical". But as a lot of l10n_fr_* modules do not have the dependency, I don't think we need it.
In theory, we could put an auto-install if mis_builder and l10n_fr are installed, but I'm not sure if we want that.

pedrobaeza: The dependency is not put due to all the `l10n_fr_oca` fork... This is not usable without a French CoA, so the dependency is real, but let's not force it.

PR #786, opened 2026-06-22
[ADD] l10n_fr_account_edi_ubl_cii : new module

remi-filament:
The current Odoo implementation with one single module l10n_fr_pdp merged end of May covers 2 things :
 * The updates of UBL / Factur-x schemas to get compliant XML files for France
 * All the logic for implementing interface with Odoo Accredited Platform (or PA for Plateforme Agréée in French)

This PR extracts the first bullet point above in a separate module so that we can have valid XML files in France even without connecting to Odoo PA.

Most of our customers in France will either :

-  Use Odoo PA
-  Use their own PA
-  Use their accounting firm PA

Except for case 1, it is not necessary to install l10n_fr_pdp (nor its dependencies account_peppol_* and auth_totp_*)

This module allows to have valid XML files for France without the burden of connecting to Odoo IAP server and extra dependencies from l10n_fr_pdp.

The split was proposed upstream but may not happen any time soon, see https://github.com/odoo/odoo/pull/268822


petrus-v (approved):
Code looks good to me

petrus-v (approved):

Hugo-Trentesaux (approved):

PR #787, opened 2026-07-02
[19.0][FIX] l10n_fr_pos_caisse_ap_ip: robustness of the payment terminal exchange + translations #1

pasgou:
Fixes several defects found during a code review of the module, plus translation updates.

### Python fixes (models/pos_payment_method.py)

- fr_caisse_ap_ip_send_payment() did not check the return value of _fr_caisse_ap_ip_prepare_message(), which can be False (unsupported currency) or an error dict (null or over-maximum amount). Both cases crashed with a 500 RPC error — the POS displayed the misleading "No answer from the payment terminal in the given time" instead of the prepared error message.
- The answer of the terminal was read with a single recv(1024): answers larger than the buffer or split across TCP packets were silently truncated. The answer is now read in a loop until the message is complete. As the Caisse-AP protocol has no length header nor end-of-message marker, when the buffer parses as a whole number of fields the socket is drained with a short timeout so that a split falling exactly on a field boundary is not mistaken for the end of the answer. A global deadline and a maximum answer size protect against a broken terminal streaming forever.
- sendall() instead of send() to avoid partial writes.
- A malformed or truncated answer made int() raise an unhandled ValueError surfacing as a 500 RPC error. The parser now raises a clear error and the caller converts it into a translated payment-issue message shown to the cashier.
- The CB tag (amount) was sent unpadded: the zfill(2) result was never written back to msg_dict, so amounts under 10 cents were sent with a single digit while the protocol requires a minimum of 2 chars.
- The "required tag missing" error message interpolated the (absent) tag value instead of the tag name, always displaying "tag None".
- The comparison of the CB tag between request and answer now strips leading zeros on both sides.
- Replaced assert with explicit exceptions in _fr_caisse_ap_ip_prepare_msg() so the checks survive python -O.

### Translation updates (i18n/)

- Added the two new user-facing error messages to the POT file, translated in French and Spanish.
- fr: the "Check" payment mode was mistranslated "Vérifier" instead of "Chèque" (the Spanish file correctly says "Cheque").
- fr + es: translated the connection-failure message (the most common runtime error — terminal off/unreachable) which had an empty msgstr.
- fr: translated the tag-mismatch message variant that had an empty msgstr; es: fixed a half-English entry.

### Testing
The pure protocol logic (serialization, TLV parsing, completeness detection, echo check) was exercised with standalone tests, and the new socket read loop was tested against a fake terminal covering: one-shot answer with connection kept open, answer split exactly on a field boundary, answer split mid-field, terminal closing after send, endless garbage stream, silent terminal.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/l10n_fr_pos_caisse_ap_ip/i18n/es.po
+++ b/l10n_fr_pos_caisse_ap_ip/i18n/es.po
@@ -53,8 +53,8 @@ msgid ""
 "Caisse AP IP protocol: tag %s is required but it is not present in the "
 "answer from the terminal. This should never happen!"
 msgstr ""
-"Caisse AP IP protocol: tag %s is required but it is not present in the "
-"answer from the terminal. ¡Esto no debería ocurrir nunca!"
+"Protocolo Caisse AP IP: la etiqueta %s es obligatoria pero no está presente "
+"en la respuesta del terminal. ¡Esto no debería ocurrir nunca!"
 
 #. module: l10n_fr_pos_caisse_ap_ip
 #. odoo-python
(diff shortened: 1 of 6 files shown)


pasgou:
Tout est corrigé, laissé à votre appréciation
@alexis-via

PR #791, opened 2026-07-27
[16.0][IMP] l10n_fr_siret: add get methods for siren/siret/nic

alexis-via:
add field is_france_country
Add consistency check between VAT and SIREN
Add hack to make SUPER PDP sandbox work

This is a full backport of all the changes introduced in l10n_fr_siret v18 for the RFE


njeudy (approved):
great ! ok for me

PR #793, opened 2026-08-07
[18.0][IMP] l10n_fr_siret: Disable checks in test mode

qgroulard:
Because Odoo core (at least) don't always use valid siret for unit tests.

See siret with invalid checksum used in [l10n_fr_pdp](https://github.com/odoo/odoo/blob/18.0/addons/l10n_fr_pdp/tests/common.py#L80).

### DIFF
--- a/l10n_fr_siret/models/res_partner.py
+++ b/l10n_fr_siret/models/res_partner.py
@@ -73,13 +73,22 @@ class Partner(models.Model):
             else:
                 rec.siret = False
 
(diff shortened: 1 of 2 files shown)


alexis-via:
I would prefer to make a PR to update the code of l10n_fr_pdp to use valid demo/test SIRET... at least we could try!

qgroulard:
Here it is: https://github.com/odoo/odoo/pull/281991

PR #796, opened 2026-08-13
[15.0][IMP] l10n_fr_siret: add get methods for siren/siret/nic

njeudy:
Backport to 15.0 of:

- 9e662b1a `[IMP] l10n_fr_siret: add get methods for siren/siret/nic`
- 78814017 `Fix res_partner.py: self.env._ is just _ in 16.0`

Adds on `res.partner`:

- `is_france_country`, computed from `res.company._get_france_country_codes()` (which
  exists on 15.0), mirroring the company field defined by `l10n_fr`;
- `_get_siren()`, `_get_siret()`, `_get_nic()`, each with a `raise_if_none` flag,
  including the SUPER PDP sandbox exception for the two non-checksummed demo SIRENs;
- a VAT ↔ SIREN consistency check in `_check_siret` (a French VAT number ends with the
  nine digits of the SIREN), and the matching `post_install` data migration, which now
  also recovers a SIREN from a too-long or partially valid SIRET instead of dropping it.

⚠️ As on 16.0, the new constraint rejects any partner carrying an inconsistent VAT/SIREN
pair on load, demo data included. The demo data of `l10n_fr_das2` and
`l10n_fr_intrastat_product` is fixed in the same commit, exactly as upstream.

Difference from the 16.0 commit: the partner form keeps its 15.0 layout — on this
version `l10n_fr_siret` still defines `company_registry` itself, and SIREN/NIC still sit
in an `oe_edit_only` div — so only the `is_france_country` condition and the read-only
SIRET on child contacts were added, not the 16.0 field rearrangement.

Why: the EN16931 / Factur-X stack reads the issuer's and customer's SIREN through these
helpers (BT-30, BT-47). This is the 15.0 half of a backport of the French e-invoicing
stack.

### DIFF
--- a/l10n_fr_das2/demo/demo.xml
+++ b/l10n_fr_das2/demo/demo.xml
@@ -17,8 +17,8 @@
     has child contacts, it will sync the siren with the childs
     while keeping the nic, so the SIRET will become invalid on the childs -->
 <record id="base.main_partner" model="res.partner">
-    <field name="siren">999999998</field>
-    <field name="nic">00019</field>
+    <field name="siren">441019213</field>
+    <field name="nic">00012</field>
 </record>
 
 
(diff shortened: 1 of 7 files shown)


njeudy:
@alexis-via backport v15.0 started !

alexis-via (approved):

PR #800, opened 2026-08-19
[18.0][IMP] l10n_fr_account_invoice_import_facturx + l10n_fr_business_document_import

alexis-via:
Improve matching of partner upon invoice import:
- in l10n_fr_account_invoice_import_facturx: update code to newer Factur-X syntax
- in l10n_fr_business_document_import: find a partner from SIREN if only VAT is set in Odoo


njeudy (approved):
ok for me :)

PR #801, opened 2026-08-19
[16.0][BACKPORT] l10n_fr_account_invoice_import_facturx + l10n_fr_business_document_import

alexis-via:
This is a backport from 18 to 16 of https://github.com/OCA/l10n-france/pull/800

Improve matching of partner upon invoice import:

- in l10n_fr_account_invoice_import_facturx: update code to newer Factur-X syntax
- in l10n_fr_business_document_import: find a partner from SIREN if only VAT is set in Odoo


njeudy (approved):
Same as 18.0 one :)

PR #803, opened 2026-09-21
[18.0][FIX] l10n_fr_account_tax_unece: add a default unce_vatex_id on exonerated VAT taxes for sale

nayatec:
As the the unece_vatex_id is mandatory for sales VAT taxes of type "E" and is now checked by new constrains on account_tax_unece this PR adds a default "NR" vatex to allow demo datas to load.

This is a proposal, that suits my immediate need but may not be the better implementation though.


alexis-via:
I prefer to set the exempt tax as inactive (it's what I did in l10n_fr_account_oca) and the python constrain doesn't apply to inactive taxes.

